PR #4, opened 2017-07-26
[WIP] Implement drmModeSetPlane

Drakulix:
A small function just to get a hang of it.

Some questions:

- Is the handle the correct place to implement `ioctl`s setting the state of a resource?
- Where can I find some docs on `drm_mode.h`? Related:
    - Where can I find out, what `flags` actually means? I think this should be a bitmap of some kind, so I would like to use `bitflags` for that instead of an `u32`
    - Where can I find out, what the return value means? I currently think it is a `bool`, but I just ignore it.
- Do you have some types / a place for common types in regards to graphics. I think something for points and rectangles, instead of this tuple madness I am currently doing, would be nice, but I was not sure, what your current approach on that is.

### DIFF
--- a/src/control/plane.rs
+++ b/src/control/plane.rs
@@ -2,7 +2,7 @@
 //!
 //! A plane is an object you can attach framebuffers to for use in displays.
 
-use control::{self, ResourceHandle, ResourceInfo};
+use control::{self, ResourceHandle, ResourceInfo, crtc, framebuffer};
 use result::*;
 use ffi;
 
@@ -71,6 +71,31 @@ impl ResourceInfo for Info {
     }
 
     fn handle(&self) -> Self::Handle { self.handle }
+}
+
+impl Handle {
+    pub fn set<T>(&self, device: &T, crtc: crtc::Handle, framebuffer: framebuffer::Handle, flags: u32, crtc_rect:((i32, i32), (u32, u32)), src_rect: ((u32, u32), (u32, u32))) -> Result<()>
+        where T: control::Device {
+
+        let mut raw : ffi::drm_mode_set_plane = Default::default();
+
+        raw.plane_id = self.as_raw();
+        raw.crtc_id = crtc.as_raw();
+        raw.fb_id = framebuffer.as_raw();
+        raw.flags = flags;
+        raw.crtc_x = (crtc_rect.0).0;
+        raw.crtc_y = (crtc_rect.0).1;
+        raw.crtc_w = (crtc_rect.1).0;
+        raw.crtc_h = (crtc_rect.1).1;
+        raw.src_x = (src_rect.0).0;
+        raw.src_y = (src_rect.0).1;
+        raw.src_w = (src_rect.1).0;
+        raw.src_h = (src_rect.1).1;
+
+        unsafe { ffi::ioctl_mode_setplane(device.as_raw_fd(), &mut raw)?; }
+
+        Ok(())
+    }
 }
 
 impl ::std::fmt::Debug for Handle {


Slabity:
> Is the handle the correct place to implement ioctls setting the state of a resource?

Honestly I'm not even sure. Setting a CRTC is done in `control::Card` and creating a framebuffer is done in `framebuffer::Info`.

We'll probably need to decide on a place to put those sort of functions.

> Where can I find out, what flags actually means? I think this should be a bitmap of some kind, so I would like to use bitflags for that instead of an u32

Unfortunately, if you can't find any documentation or examples for those flags, then you'll probably need to look through the kernel code to see what it's checking for.

> Where can I find out, what the return value means? I currently think it is a bool, but I just ignore it.

We just return an error if there is any, otherwise an `Ok(())` value is fine.

> Do you have some types / a place for common types in regards to graphics. I think something for points and rectangles, instead of this tuple madness I am currently doing, would be nice, but I was not sure, what your current approach on that is.

Right now I just use tuples. Optimally we would use an external crate that provides those datatypes though.

Drakulix:
> We'll probably need to decide on a place to put those sort of functions.

I think that depends, if you want to follow the drm_mode.h api closely or if you want to try out new approaches?

The former has the benefit for users of the C library, that it would already feel familiar. It also would theoretically make it possible to write some C-bindings for this crate to make it a better documentated alternative to mesa's drm_mode.

A fresh start however might result in a better API. In that case do you think that this function is currently in a good place?

> Unfortunately, if you can't find any documentation or examples for those flags, then you'll probably need to look through the kernel code to see what it's checking for.

Alright, I will see what I can find out before merging.

> Right now I just use tuples. Optimally we would use an external crate that provides those datatypes though.

Is this something, that needs to be handled by another crate?
Most of those implementation I have seen just do something like:

`type Point = (i32, i32);`

Which is compatible to all other crates by design, as it is just an alias, but that would make the api far more readable.

Slabity:
> I think that depends, if you want to follow the drm_mode.h api closely or if you want to try out new approaches?

I'm not against another approach, but really the only concern I have is how we handle methods that change the state of the device. Reading from the device is pretty easy.

Any ideas?

> Is this something, that needs to be handled by another crate?

It's not necessary, but if there's a widely used crate that already provides a more expressive interface, then I would much prefer it. For now, creating a basic `Point` and `Size` type would be sufficient.

Drakulix:
> I'm not against another approach, but really the only concern I have is how we handle methods that change the state of the device. Reading from the device is pretty easy.
>
> Any ideas?

What is your concern in regards to that? I would just handle the device as a parameter, that I pass into the function of these `Handle`s, like the framebuffer's `Info` for creating a new one.

Is a framebuffer fixed to a single device? If yes, just store the device file descriptor in all `Info`s and `Handle`s, so it is not required anymore as a parameter and fixed as well.

Are there any actual race conditions with ioctl calls? If yes just mark the device and all structs storing the device file descriptor as `!Sync`.
As long as the device is not accessible from many threads, there can only be one ioctl call at once, meaning we can just pass in the `Device` as a read-only parameter for those calls or store it, if it is fixed.

Of course with a single FD the user can create multiple devices, but that would also not be save for your current approach, so it should be documented, that one open device should be unique.

Slabity:
> What is your concern in regards to that? I would just handle the device as a parameter, that I pass into the function of these `Handle`s, like the framebuffer's `Info` for creating a new one.

That's probably how it should be done. I'm not really sure whether it should be part of `Handle` or `Info` though. Or maybe just make the function not a part of any particular object. Just having functions like `control::framebuffer::create(...)`, `control::crtc::set(...)`, and `control::plane::set(...)` would be very easy to do.

> Is a framebuffer fixed to a single device? If yes, just store the device file descriptor in all `Info`s and `Handle`s, so it is not required anymore as a parameter and fixed as well.

I tried this before, but unfortunately it meant keeping track of some very complicated lifetimes and even more complicated generics. Reference counting the device is an option, but that brings up even more complications.

Drakulix:
> That's probably how it should be done. I'm not really sure whether it should be part of Handle or Info though. Or maybe just make the function not a part of any particular object. Just having functions like control::framebuffer::create(...), control::crtc::set(...), and control::plane::set(...) would be very easy to do.

I like that approach.

> I tried this before, but unfortunately it meant keeping track of some very complicated lifetimes and even more complicated generics. Reference counting the device is an option, but that brings up even more complications.

Yes, that is why I suggested just storing the `RawFd`. But both solutions work fine.

Slabity:
> Yes, that is why I suggested just storing the RawFd. But both solutions work fine.

If you store the raw file descriptor without tracking the lifetime of the file, then you can end up with some pretty nasty behavior. For example, if you close the device and open up something else, then all of the sudden that file descriptor stored in your `framebuffer::Handle` is now pointing to some random file.

Drakulix:
Yes that is right. You would need to ref-count the wrapped RawFd internal and consume it on device creation, so you have full control, over when it is dropped. But I like the current approach, as it is far cleaner and easier to implement.

PR #7, opened 2017-08-05
Add crtc gamma control

Drakulix:
Added functions akin to `drmModeCrtcGetGamma` and `drmModeCrtcSetGamma` . I had to introduce a new error type to make a rusty-api, but now, I get some compiler warnings.
I have not yet used the `error-chain` crate (although it seems very nice), maybe I don't understand, how I should do it?

The warning is as follows:
```
warning: doc comment not used by rustdoc
  --> src/result.rs:3:1
   |
3  | / error_chain! {
4  | |     foreign_links {
5  | |         Unix(nix::Error);
6  | |     }
...  |
13 | |     }
14 | | }
   | |_^
   |
   = note: #[warn(unused_doc_comment)] on by default
   = note: this error originates in a macro outside of the current crate
```

Other then that, this is ready for merge in my eyes. What do you think @Slabity ?

### DIFF
--- a/src/control/crtc.rs
+++ b/src/control/crtc.rs
@@ -322,6 +322,67 @@ pub fn move_cursor<T>(device: &T, handle: Handle, to: iPoint) -> Result<()>
     Ok(())
 }
 
+#[derive(Debug, Clone)]
+pub struct GammaRamp {
+    pub red: Box<[u16]>,
+    pub green: Box<[u16]>,
+    pub blue: Box<[u16]>,
+}
+
+pub fn gamma<T>(device: &T, handle: Handle) -> Result<GammaRamp>
+    where T: control::Device {
+
+    let info = Info::load_from_device(device, handle)?;
+
+    let mut raw: ffi::drm_mode_crtc_lut = Default::default();
+    raw.crtc_id = handle.as_raw();
+    raw.gamma_size = info.gamma_length;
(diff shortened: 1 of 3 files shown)


Slabity:
In the `gamma` function, you call `Vec::with_capacity` a few times. This allocates memory, but it does not initialize that memory or set the length of the `Vec`. A call to `into_boxed_slice` will drop this excess memory, and the length will always be zero.

I've created a macro to assist in these situations called `ffi_buf`. It does all the heavy lifting, and all you need to do is provide the pointer location and length. I would recommend taking a look at `connector::load_from_device` to see how it works.

Drakulix:
Thanks. I will take the time to fix this soon.

Drakulix:
Should be done now.

Slabity:
Your `set_gamma` function uses the wrong ioctl. Can you write a test for this?

Drakulix:
I have no actual idea, how these values work to write a test for them, but I will figure out and fixup that ioctl.

Slabity:
In that case, let's save these gamma control functions for later. I'll look into how it works.

Drakulix:
Fixed, Tested and rebased onto master. You should see a nice red with these values. There is nothing special about it.

Slabity:
That was fast. Looks like it works on my machine as well.

PR #8, opened 2017-08-15
[WIP] drm_mode_add_fb2

Drakulix:
This is quite a though one, that why I am presenting this PR so unpolished and complete.

This tries to use `drm_mode_add_fb2`, when supported by the underlying system and falls back to the legacy ioctl otherwise.

To do this, it introduces a `PixelFormat`, which can be converted back to `depth` and `bpp` if supported. (Note that this is a superset. There is no valid depth and bpp combination, that is not represented by a valid  pixel format for `drm_mode_add_fb2` as far as I know.)

The pixel format values are coming from `libdrm/drm_fourcc.h`, which is why it has that weird name in the sys-crate, to follow the libdrm api closely. Bindgen fails on that one and I already had coded an [almost identical file](https://github.com/Drakulix/gbm.rs/blob/master/src/lib.rs#L67) for my `gbm` crate (which I still need to move here at some point), so I just did use that, it closely resembles the original file. (libgbm and libdrm share the possible formats and their representation.)

This conversion to depth and bpp is necessary for creating dumb buffers and for falling back to the old ioctl and throws an error if not supported for the given pixel format.

This gives a nice API (in my opinion), but creates a range of problems:
- The Framebuffer `Info` struct cannot contain the format, because there is no `drm_mode_get_fb2`-ioctl to fixup `load_from_device`. (Google actually wants to patch this, but this means if `drm_mode_add_fb2` is available there will still be the possibility that `drm_mode_get_fb2` is not available on older kernel versions, so this only makes matters more complicated.)
- The Framebuffer `Info` struct might even contain invalid information. Some pixel formats exclusive to `drm_mode_add_fb2` can cause wrong depth and bpp values.
- Adding multiple buffers, flags and modifiers will make the function signature more complicated and introduce more errors, that can happen. E.g. if multiple buffers are set, but we need to fallback, which cannot be done with multiple buffers, we need to catch that and error. At least this is possible, but makes the fallback rather useless, as it only works in very simple cases and breaks almost immediately for any other case.
- Fixing this by creating a new function like `create2` that uses the new ioctl exclusively and keep `create` is also not optimal. It can still throw an `UnsupportedPixelFormat` error, which is more difficult then the old api. But we do not want to introduce multiple `Buffer` traits to restore the old api (or do we?). 

I do view the current solution as the best, I can think of. But I am not completely happy with it and would like to have some feedback on this @Slabity. Thanks :)

### DIFF
--- a/drm-sys/src/fourcc.rs
+++ b/drm-sys/src/fourcc.rs
@@ -1,0 +1,75 @@
+macro_rules! __fourcc_code {
+    ($a:expr, $b:expr, $c:expr, $d:expr) => (
+        ($a as u32) | (($b as u32) << 8) | (($c as u32) << 16) | (($d as u32) << 24)
+    )
+}
+
+/* color index */
+pub const DRM_FORMAT_C8 	:u32 = __fourcc_code!('C', '8', ' ', ' '); /* [7:0] C */
+
+/* 8 bpp Red */
+pub const DRM_FORMAT_R8		:u32 = __fourcc_code!('R', '8', ' ', ' '); /* [7:0] R */
+
+/* 16 bpp RG */
+pub const DRM_FORMAT_GR88		:u32 = __fourcc_code!('G', 'R', '8', '8'); /* [15:0] G:R 8:8 little endian */
+
+/* 8 bpp RGB */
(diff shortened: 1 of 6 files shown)


Drakulix:
@Slabity Did you already found the time to look into this one?

Slabity:
Hey, sorry. I was trying to see if the `fourcc` crate would be a better fit, but it's apparently deprecated.

This does make more sense than having separate `bpp` and `depth` values.

PR #10, opened 2017-08-17
Feature/page flip

Drakulix:
Depends on #9 for cleanup (necessary for chaining the tests reliably).

Implements page flipping and adds a (rather obnoxious looking) test that uses it.
Also implements dump buffer destruction for cleanup.

### DIFF
--- a/drm-sys/src/platforms/linux/x86_64/bindings.rs
+++ b/drm-sys/src/platforms/linux/x86_64/bindings.rs
@@ -4071,7 +4071,7 @@ pub struct drm_event_vblank {
     pub tv_sec: __u32,
     pub tv_usec: __u32,
     pub sequence: __u32,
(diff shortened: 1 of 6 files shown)


Slabity:
I am hesitant to make manual changes to `drm-sys` bindings. These are automatically generated by the `build.rs` script, and changing them manually will result in differing implementations if someone needs to use bindgen. The bindings are only there for convenience and to remove dependencies on libdrm's header files.

I'm still merging this, but I feel this might break the `use-bindgen` feature. I'll look into setting up CI to test and prevent these issues.

Drakulix:
This change is generated with newest mesa headers. I think `drm-sys` should include the header file it is generating from.

Slabity:
Ah, that makes sense. Everything is fine then.

PR #11, opened 2017-08-31
Remove associated_consts feature

Drakulix:
Associated consts is stabilized in the current form used by drm-rs, already works in beta and will be in rust stable 1.20. (see https://github.com/rust-lang/rust/issues/29646#issuecomment-325419397)

Using the feature flag prevents compilation on beta, as the feature flag is generally not allowed on stable and beta channels even if the feature in question is already stabilized.

tl;dr This fixes compilation on current beta and upcoming stable 1.20

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,3 @@
-#![feature(associated_consts)]
-
 #[macro_use]
 extern crate nix;
 extern crate drm_sys;


Drakulix:
Could you release 0.2.1 with these changes? :)

Slabity:
Sure thing! I'll release it later today once the 1.20 compiler gets officially released.

Drakulix:
Ping

Slabity:
The version on crates.io is 0.2.1 now, I just haven't committed it to github.

PR #12, opened 2017-09-13
Fix page_flip userdata

Drakulix:
Sadly the userdata code of my PR #10 is flawed, because a raw pointer of a TraitObject (`*mut Any` from `Box<Any>`) is actually whats called a "fat" pointer in rust, meaning it is 128 bit wide. Of course that does not fit into a `u64` and so downcasting the `Box<Any>` does never work, but always returns an error.

This PR fixes this be wrapping the `Box<Any>` into a tuple struct, that can be casted into a pointer normally before passing it into the ioctl. This all happens transparently, there are no visible api changes in this PR.

(I also took the time to fix some warning, `Some(handler)` is now a `& &mut Handler` instead of the previous `&mut &mut Handler`.)

Would be cool, if this could be released as 0.2.2 soon, I am working with a git dependency in the meantime, as I require this feature.

### DIFF
--- a/src/control/crtc.rs
+++ b/src/control/crtc.rs
@@ -130,6 +130,8 @@ pub enum PageFlipFlags {
     PageFlipAsync = ffi::DRM_MODE_PAGE_FLIP_ASYNC,
 }
 
+struct FatPtrWrapper(Box<Any>);
+
 pub fn page_flip<T, U>(device: &T, handle: Handle, fb: FBHandle, flags: &[PageFlipFlags], userdata: U) -> Result<()>
     where T: control::Device, U: 'static {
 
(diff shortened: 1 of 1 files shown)


vberger:
Slight question, is there a reason why the `user_data` field here seems to be of type `u64` rather than `*mut c_void`, like traditionally in C programs ?

This could cause issues with 32bits platforms for example, couldn't it?

Drakulix:
The u64 is generated by bindgen, I have no seen anything else in any drm header. I assume the ioctl data structure is the same across all architectures.

Slabity:
@vberger The DRM API expects all pointers to be 64-bit regardless of the architecture. I'm not sure the reasoning behind it though.

PR #13, opened 2017-09-15
Feature/encoder

Drakulix:
I need functionality to check which encoders are supported by a crtc and which connectors support which encoders. (see https://github.com/Smithay/smithay/pull/34/commits/75601d8f6a0633fd9e5d073d80c44979dc098047)

Also there seems to be no reason not to let the `Handle`s implement `std::hash::Hash`, which is really useful when filtering for encoders (as seen in the commit above), so that was added as well.

I have seen you already had different plant with your `CrtcListFilter`, but I don't understand where you were going with it and this was quick and easy to implement. If you wish for some other solution, please tell me and I will adjust this PR until it fits your vision and my needs.

https://github.com/Smithay/smithay/pull/34 depends on this PR, so just like with #12, I would :heart: a quick release, if possible.

### DIFF
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -18,7 +18,7 @@ use result::*;
 /// [`ResourceHandle`]: ../ResourceHandle.t.html
 /// [`connector::Info`]: Info.t.html
 /// [`ResourceHandles::connectors`]: ../ResourceHandles.t.html#method.connectors
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Handle(control::RawHandle);
 
 /// A [`ResourceInfo`] for a connector.
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -89,6 +89,11 @@ impl Info {
     pub fn modes<'a>(&'a self) -> &'a [control::Mode] {
         &self.modes
     }
+
+    /// Returns a list containing each supported [`encoder::Handle`].
+    pub fn encoders<'a>(&'a self) -> &'a [control::encoder::Handle] {
+        &self.encoders
+    }
 }
 
 impl ResourceHandle for Handle {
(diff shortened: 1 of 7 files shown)


Drakulix:
Rebased after merge of #12.

Slabity:
Could you please explain what this line of code is doing?

```rust
self.possible_clones & (Wrapping(1u32) << crtc.as_raw() as usize).0 != 0
```

Drakulix:
Sure.

The line is taken (and rustified) from [this excellent modesetting-howto](https://github.com/dvdhrm/docs/blob/fad7c3203b14e67053e0fc41d8490138b8ff47dd/drm-howto/modeset.c#L386)

After wondering how to works, if possible_crtcs is really a bitflag and all of that, I found [this comment](https://dvdhrm.wordpress.com/2012/09/13/linux-drm-mode-setting-api/#comment-215) under the blog post of said howto.

Since I did not find any more related resources about how to use the possible_crtc's value and that comment shows, that the code should be valid, I just went with it. And so far I have not found a single bug in it.

Slabity:
Correct me if I'm wrong, but you're shifting `1` by the actual ID of the CRTC, while that modesetting code is shifting it by the index of the CRTC within the set of resources.

If my interpretation is correct, then if you have a CRTC with an ID of `42`, then you're doing `possible_clones & (1 << 42)` which would always be zero.

Drakulix:
Oh my, you are right, sorry.

Do you have a proposed solution for this? I kinda need that feature and it would be nice to have it implemented rather sooner then later and I would happily do that.

Slabity:
My plan with `CrtcListFilter` was to allow it to be 'applied' to the resources structure and return an iterator over the IDs that were possible.

I've just gotten some extra time to do this though. I'll see if I can get it working tonight.

Drakulix:
> EDIT: Just confirmed on my machine that this doesn't work, so unfortunately I had to revert this. I'll cherry pick the commit that adds the hash trait though, as there's really no reason not to include that.

Quick heads-up this landed on master anyway, so master appears to be broken right now (well it works, it just does not filter).

Drakulix:
> I've just gotten some extra time to do this though. I'll see if I can get it working tonight.

Is this still on your list or should I try to tackle it?

Slabity:
Sorry, I just pushed a way to get a `CrtcListFilter` from an encoder. You can use it on `ResourceHandles` to get a buffer of compatible CRTCs.

I know it's not the most sophisticated way, but it works and should let you use it in Smithay for now. In the future I'd like to change it to some sort of zero-overhead iterator, but that's not necessary right now.

PR #14, opened 2017-09-19
Add a lot of docs

Drakulix:
Alright, so I did a very poor job at documentation any of my changes so I decided to start just documenting the whole crate.

Of course this is quite difficult with a lot of the original drm api not being well documentated and me having not a good idea of how everything works, but it is a start.

I will maybe continue to work on this PR and also would welcome anyone to contribute, but I am also happy with a merge in this state, I can always follow up with more.

Also feedback is very welcome.

Some *API changes* creeped into this, if I should move these to a separated PR @Slabity just tell me, they are really small:

- `set_cursor` and `set_cursor2` now take a `Buffer` reference instead of a separate `Buffer::Id` and `dimensions`.
- `crtc::page_flip` does not take a version anymore as this is not necessary at all.
- `DumpBuffer::map` now takes a mutable reference, as otherwise on some drivers you can get multiple mutable references to the same memory area.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -4,13 +4,16 @@ use ffi::fourcc::*;
 pub type RawId = u32;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// GEM handle of a buffer
 pub struct Id(RawId);
 
 impl Id {
+    /// Convert into a GEM handle from the raw type
     pub fn from_raw(raw: RawId) -> Self {
         Id(raw)
     }
 
+    /// Convert into the raw type
     pub fn as_raw(&self) -> RawId {
         self.0
     }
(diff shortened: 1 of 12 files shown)


Slabity:
Hey @Drakulix, it looks like the vast majority of the changes were made by `rustfmt`, correct? While a great idea, I'd prefer them to be in a separate from any documentation or API changes just to make it easier to find potential bugs.

Slightly related, what do you think about a linting program like `clippy`?

Drakulix:
@Slabity Sorry that must have been my IDE (trying out RLS). I will try to revert all rustfmt related changes.

clippy is very good, but I usually only use it to lint my code periodically and fix a bunch of stuff, because it breaks so frequently with new rust-nightly builds, that I find it hard to automate clippy testing.

Drakulix:
@Slabity Fixed and rebased on current master.

I also was able to separate one api change into another commit, the other two were not so easy, so I did not bother, but of course I can do those as well, if you want.

But all rustfmt changes should not be gone.

Drakulix:
Bad typo. I meant those rustfmt changes should now(!) be gone.

Slabity:
Looks good to me!

PR #15, opened 2017-09-20
Allow event handlers to use the device mutably

Drakulix:
Pretty much what the title says.

If you don't like that using a mutable device is now a requirement, I can add an additional `handle_device_mut` function and handlers and leave the current one untouched.

### DIFF
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -18,7 +18,7 @@ use result::*;
 /// [`ResourceHandle`]: ../ResourceHandle.t.html
 /// [`connector::Info`]: Info.t.html
 /// [`ResourceHandles::connectors`]: ../ResourceHandles.t.html#method.connectors
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Handle(control::RawHandle);
 
 /// A [`ResourceInfo`] for a connector.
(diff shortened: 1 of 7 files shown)


Slabity:
Just wondering, but why does the device need to be mutable?

Drakulix:
The struct I implement your Device trait for has some other functions requiring a mutable reference, that I would like to use inside the handler. There is no reason not to allow that and I would like not to use a Mutex, if not necessary. My device struct is quite complex (not just a `File` wrapper) and there is no easy way around that. (Take a look at `src/backend/drm/mod.rs` in smithay, if you are interested.)

Drakulix:
I thought exactly the same about that after I did that reimplementation. But I wanted an api true to the original to make it feel more familar for users and because I did not have a good idea about libdrm at that time. I thought there must have been a reason, that it was designed that way, which is why I later removed that version parameter and now made the device mutable, when I really should have taken the time to refactor it.
I will change that tomorrow.

Slabity:
My guess is that they planned out the API to be asynchronous, but dropped it for some reason and kept the API around.

I'm not too interested in keeping the mistakes that libdrm made though. How about a simple iterator owned by the device that returns events? That would probably make it way nicer to use.

vberger:
> What the hell? Do they not trust user programs to be able to call their own handler correctly? You shouldn't be forced to let the library call your own function for you, especially when you need to call their function to call your own function.

Funnily enough, this is exactly the same thing with wayland. (Though they added some ways for binding to plug themselves slightly lower level, with a generic callback to dispatch messages.)

Anyway, if the API can reasonably be converted to event iterators, I can confirm this is likely the best route. 👍 (it wasn't really possible for wayland, as half the arguments of the callbacks were references...)

Drakulix:
See #16

PR #16, opened 2017-09-27
[WIP] Rework handle_event into an event iterator

Drakulix:
Ref #15.

This is untested, so I do not feel comfortably merging, but I wanted to get a quick feedback from you @Slabity, if this is going into the right direction from your perspective.

This is basically how everything worked before just wrapped into an iterator, so I am quite confident, that it will most likely work, but I noticed a hole in my earlier implementation.

If there are more events then fit into 1024 bytes, they will be thrown away (which should not happen usually, but is kinda bad, I took that value from libdrm, so most programs today seem to be fine with it). I think we can do better, but it is hard in rust, as there is no way to check how much bytes I can read without blocking the device. I basically need to know, how many bytes are available and then read exactly those.

Do you have a good idea how that could be done? I think the FIONREAD ioctl should make this possible, so I am thinking about adding that to src/ffi.rs. Thoughts?

### DIFF
--- a/src/control/crtc.rs
+++ b/src/control/crtc.rs
@@ -126,21 +126,19 @@ pub enum PageFlipFlags {
     PageFlipAsync = ffi::DRM_MODE_PAGE_FLIP_ASYNC,
 }
 
-struct FatPtrWrapper(Box<Any>);
-
 /// Queue a page flip on the given crtc.
 ///
 /// On the next vblank the given framebuffer will be attached to the
 /// crtc and an event will be triggered by the device, which will be indicated by it's fd becoming
 /// readable. The event can be received using `handle_event`.
-pub fn page_flip<T, U>(device: &T, handle: Handle, fb: FBHandle, flags: &[PageFlipFlags], userdata: U) -> Result<()>
-    where T: control::Device, U: 'static {
+pub fn page_flip<T>(device: &T, handle: Handle, fb: FBHandle, flags: &[PageFlipFlags]) -> Result<()>
+    where T: control::Device {
 
     let mut raw: ffi::drm_mode_crtc_page_flip = Default::default();
     raw.fb_id = fb.as_raw();
     raw.crtc_id = handle.as_raw();
     raw.flags = flags.into_iter().fold(0, |val, flag| val | *flag as u32);
-    raw.user_data = Box::into_raw(Box::new(FatPtrWrapper(Box::new(userdata) as Box<Any>))) as u64;
+    raw.user_data = handle.as_raw() as u64;
 
     unsafe {
         try!(ffi::ioctl_mode_page_flip(device.as_raw_fd(), &mut raw));
@@ -172,8 +170,8 @@ pub struct VblankEvent {
     pub frame: u32,
     /// duration between events
     pub duration: Duration,
-    /// userdata as passed into `page_flip`
-    pub userdata: Box<Any>,
+    /// crtc that did throw the event
+    pub crtc: Handle,
 }
 
 /// Page Flip event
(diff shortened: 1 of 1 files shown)


Slabity:
That's funny. I was just finishing up a different implementation of the event iterator. Though yours is a bit cleaner and doesn't require ownership of the device.

There should be some sort of `Unknown` type in `Event` with just a byte slice for its content. That way we don't `_ => continue` on arbitrary data (in case the user wants to see it for some reason). That would also allow you to replace the loop entirely with something like `if self.i < self.amount { ... } else { None }`.

Also I think you should just call the event structs `Events` and `Event`, but it doesn't really matter right now.

Drakulix:
> Though yours is a bit cleaner and doesn't require ownership of the device.

Ownership would defeat the point for me, because - again - I would not be able to use the device mutably during event handling.

> There should be some sort of Unknown type in Event with just a byte slice for its content.

Sure thing, sounds better then my hack.

> Also I think you should just call the event structs Events and Event, but it doesn't really matter right now.

Easy. Will fix that right away.

> EDIT: Also might be useful to take the RawFd and temporarily transform it into something that already implements Read rather than implement it ourselves. Maybe even use BufReader.

I could use a `File` because that implements `IntoRawFd`. We cannot let the object close the file descriptor on destruction, which is why I though my current implementation is cleaner, because you cannot forget to destruct it again probably, so no closure happens. `BufReader` does not seem to allow this at all.
I will keep it like it works right now, but if you dislike it, I would change it into a `File`, no problem.

Slabity:
Also, since we aren't doing the callback method, is there any point in having the `user_data` fields? The DRM API doesn't do anything with it, and I'd assume the user will keep track of their data in a way they'd want.

Drakulix:
Yes there is a point, because `page_flip` - where the userdata originates from - is asynchronous. On drivers not supporting the `crtc_id` parameter for example there is no way to know for which crtc the page flip was queued, when the event comes back. If I drive multiple crtc's at the same time, this information is crucial, so I use the userdata parameter in smithay to attach the `crtc::Handle` to it.

Slabity:
That makes sense. But is there a reason to use arbitrary data? I can't think of any information other than a `crtc::Handle` that would be useful here. We could have `page_flip` be defined as:

```rust
pub fn page_flip<T>(device: &T, handle: Handle, fb: FBHandle, flags: &[PageFlipFlags]) -> Result<()>
    where T: control::Device {

    let mut raw: ffi::drm_mode_crtc_page_flip = Default::default();
    raw.fb_id = fb.as_raw();
    raw.crtc_id = handle.as_raw();
    raw.flags = flags.into_iter().fold(0, |val, flag| val | *flag as u32);
    raw.user_data = handle.as_raw() as u64;

    unsafe {
        try!(ffi::ioctl_mode_page_flip(device.as_raw_fd(), &mut raw));
    }

    Ok(())
}
```

And then that would allow the user to control how and where the data is located by keeping track of the `Handle`.

Drakulix:
Well, I guess you are right on that one. I will change it and we can keep it until someone actually requires arbitrary data and opens an issue about it.

Drakulix:
I let the if clause regarding the crtc_id field of the vblank_event in there, to document, that we could use the userdata field for something else on modern implementations.

Slabity:
Okay, sounds good to me!

Slabity:
And with this done, it looks like pretty much all of non-atomic modesetting is feature complete!

I'm still trying to flesh out how to organize properties for atomic operations, but until then it looks like Smithay can have a working DRM backend.

Drakulix:
Indeed! I will try to get the current master to work with smithay, if everything appears to be fine, I will ping you again in the next days, so you can release 0.2.4/0.3 on crates.io.

PR #17, opened 2017-10-25
Add prebuilt FreeBSD bindings

### LINKED ISSUE #1: Add OpenBSD and FreeBSD prebuilt bindings
Both OpenBSD and FreeBSD expose a DRM API. We should try to include prebuilt bindings for these operating systems.

myfreeweb:
Partially fixes #1. (Actually, why not always use bindgen on the fly? There's not just OS/arch differences but also different libdrm versions.)

BTW, tests are broken by 065f596918 and so on…


Slabity:
Well we don't use libdrm. We build our own library to use the DRM API directly.

To be honest, the only differences between system's we need to worry about is the magic ioctl numbers, as that can vary between systems.

PR #23, opened 2018-02-06
Impl Hash for Mode

Drakulix:
Hey @Slabity,
I have some more small changes that would improve usage of your library for me.

1. I have added a `Hash` implementation for the drm modes (to allow easy filtering for modes supported by multiple connectors via `HashSet`).
2. I have removed `name` from the `PartialEq` implementation (and not added it to `Hash`), because that seems to be connector dependent and the kernel source code suggests, that it is not used for mode setting. (I am not entirely sure about it, but my monitors report different names for otherwise identical modes and setting both mirrored using either mode works for me.)
3. I have removed a double check (possible a typo) from the `PartialEq` implementation of `clock`.

As always, if you merge this a quick crates.io release would be very nice.

Thanks!

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -1,7 +1,7 @@
 use result::*;
 use ffi;
 use std::ffi::CStr;
-use std::hash::Hash;
+use std::hash::{Hash, Hasher};
 pub mod connector;
 pub mod encoder;
 pub mod crtc;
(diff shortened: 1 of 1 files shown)


Slabity:
This has been merged and pushed to crates.io as version `0.3.2`

PR #24, opened 2018-02-08
Add getter for `vrefresh` to `Mode`

Drakulix:
Sorry to bother you again,
but it seems there is no way currently to get the refresh rate of a `Mode`.

I have added a getter for it.


Slabity:
Published as version `0.3.3`

PR #29, opened 2018-11-06
Allow `Buffer`s passed as reference to be unsized

Drakulix:
This removes the requirement of `Buffer`s to be sized, when passed as a reference.
`drm-rs` only calls into the trait methods anyway.

I know that development is happening on another branch, but I have no idea how close that is to a release. This PR would be a nice quality of life improvement for `smithay` right now, so if you do not see any problems, I would be very happy to see a 0.3.4 release.


Slabity:
Looks good to me. I'll have 0.3.4 out tonight.

PR #32, opened 2019-04-27
Fix some issues when building for ARM

emdash:
- Export the `use_bindgen` feature at the top level, so non-86x targets can enable it.
- Use the `nix` types, `c_char`, and `off_t`, which seem to be slightly different on ARM.
- Replace one instance of `mem::transmute` with plain cast.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -18,7 +18,7 @@ macro_rules! ffi_buf {
             use std::mem;
 
             let mut buf = unsafe { vec![mem::zeroed(); $sz as usize] };
-            *(&mut $ptr) = unsafe { mem::transmute(buf.as_mut_ptr()) };
+            *(&mut $ptr) = unsafe { buf.as_mut_ptr() as u64};
             buf
         }
     )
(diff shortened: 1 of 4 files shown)


src/ffi.rs
```diff
@@ -18,7 +18,7 @@ macro_rules! ffi_buf {
             use std::mem;
 
             let mut buf = unsafe { vec![mem::zeroed(); $sz as usize] };
-            *(&mut $ptr) = unsafe { mem::transmute(buf.as_mut_ptr()) };
+            *(&mut $ptr) = unsafe { buf.as_mut_ptr() as u64};
```
emdash: On 32-bit targets, this will not compile, since the destination type is 64-bits, while `buf.as_mut_ptr()` is 32-bits. If there is a way to do this that doesn't sacrifice the generality of the macro, I will happily change it. In practice, this macro is only ever used to populate 64-bit ingegers.

Slabity: Makes sense. Might not even need the unsafe at this point, but I'll look at that later.

PR #34, opened 2019-05-06
Reimplement cursor functions

Drakulix:
Building upon buffer support in #33, I quickly added the cursor functions.

This is a bit closer to the libdrm-function then the old implementation (on master). It allows `None` being passed in as a buffer, instead of providing a `clear_cursor` function. Not sure though which is better for *rusty* bindings.

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -600,14 +600,6 @@ pub trait Device: super::Device {
 
         Ok(dumb)
     }
-
-    /// Free the memory resources of a dumb buffer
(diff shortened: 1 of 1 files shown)


Drakulix:
Rebased onto #38

Drakulix:
@Slabity Any specific reason against this PR?

- I could combine the set-ioctls into a single one with an optional hotspot and automatic fallback?
- Or re-introduce clear_cursor instead of set_cursor with an optional buffer again?

Drakulix:
Rebased on current develop (99be0f9) without changes.

PR #35, opened 2019-05-07
Reimplement gamma functions

Drakulix:
Another quick reimplementation.

In #25 you mention, that you do not want to do any memory allocations in drm-rs anymore, which is why I removed the previously existing `GammaRamp` struct (containing `Box<[u16]>` color ramps) and used slices instead.

This gives us a c-style-output-argument function signature for `get_gamma`, but I was unable to come up with a better solution, since the gamma_length can vary between devices (and theoretically CRTCs of the same device).

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -382,6 +382,50 @@ pub trait Device: super::Device {
 
         Ok(prop_val_set)
     }
+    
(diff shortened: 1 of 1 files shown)


Drakulix:
Rebased onto latest develop

PR #37, opened 2019-05-09
Implement prime fd to/from handle conversion

Drakulix:
Again building on the buffer implementation in #33, allowing to convert between buffer handles and prime file descriptors.


Drakulix:
Rebased on current develop (99be0f9) without changes.

PR #36, opened 2019-05-08
Reimplement framebuffer functions

Drakulix:
Based (again) on #33 this adds the `PlanarBuffer` trait as well as `add_fb(2)`, `dirty_fb` and `rm_fb` functions.

PR #38, opened 2019-05-12
Add new buffer traits

Drakulix:
Split off commits of earlier revisions of PRs #33 and #36.
Allows for easier cherry-picking and discussion.

- [x] Move the `ResourceHandle` and `ResourceType` into the crate root to be used by the buffer module
- [x] Finish the implementation of the `Buffer` trait
- [x] Finish the implementation of the `PlanarBuffer` traits
- [x] Implement `open_buffer` and `close_buffer`
- [ ] `flink` is deprecated and not (yet) implemented
- [x] Add a `PlanarBuffer` implementation for every `Buffer`, since every Buffer is a single-plane buffer

### DIFF
--- a/src/buffer/format.rs
+++ b/src/buffer/format.rs
@@ -1,6 +1,6 @@
 //! Color formats using standard FourCC.
 
-use ffi::fourcc::*;
+use drm_ffi::fourcc::*;
 
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 #[allow(missing_docs)]
--- a/src/buffer/mod.rs
+++ b/src/buffer/mod.rs
@@ -103,3 +103,25 @@ pub trait PlanarBuffer {
     /// The offsets of the buffer.
     fn offsets(&self) -> [u32; 4];
 }
+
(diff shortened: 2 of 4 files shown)


Slabity:
I will resolve these conflicts (and the other pull requests) and merge them within the next week

Drakulix:
> I will resolve these conflicts (and the other pull requests) and merge them within the next week

Already rebased (and conflicts resolved). I also removed the commit moving the `(Raw)ResourceHandle` into the crate root, because I think `buffer::Handle/Name` are no `ResourceHandle`s as I was unable to identify their `FFI_TYPE`.

They are not just wrappers for `u32`.

Slabity:
> They are not just wrappers for u32.

Yep. The kernel subsystem shares the set of 32-bit handles between 6 (maybe 7) different resources; Connectors, encoders, CRTCs, framebuffers, planes, and properties (and maybe blobs - still investigating)

Buffer handles exist as their own thing and can overlap with the values of those above. So if you used a buffer handle and ran something like `get_properties` on it you would get the properties of some random encoder or plane or something.

PR #57, opened 2019-11-28
Fix some tiny problems in examples that prevented some of them from running

ctsrc:


Slabity:
This looks good to me. I'll merge it now.

PR #71, opened 2021-03-15
Added prebuilt bindings for Linux on aarch64 and fixed broken examples.

scrblue:
Bindings were unable to be generated on an aarch64 machine unless the following environment variable was set: `RUSTFLAGS=-Ctarget-feature=-crt-static`, but after generation could be inserted into src/platforms of drm-sys. Adjusted examples until they built and ran properly on both aarch64 and x86_64.


katyo:
@scrblue I assume the generated bindings for any 64-bit linux will be entirely same dispite of concrete architecture. (Also for any 32-bit ones)

I tested such fact by building drm-sys crate with 'use_bindgen' feature for `x86_64` and `aarch64` also for `i686`, `armv6` and `armv7`.

Drakulix:
> I tested such fact by building drm-sys crate with 'use_bindgen' feature for `x86_64` and `aarch64` also for `i686`, `armv6` and `armv7`.

Oh that is good to know. So instead we should get rid of the needless architecture separation in `drm-sys`?
@scrblue Could you change your PR to just use the provided bindings regardless of architecture and move them up into their respective system-directories? (So `linux/x86_64/bindings.rs` becomes `linux/bindings.rs`.)

katyo:
@Drakulix Note that 32-bit and 64-bit bindings is differs in some c types. But the bindings for platforms with same word size looks identical.

Drakulix:
> @Drakulix Note that 32-bit and 64-bit bindings is differs in some c types. But the bindings for platforms with same word size looks identical.

If that gets translated into the correct type (namely usize/size) this should not really be a problem, right?
Maybe this is difficult to validate and to error-prone, but it feels kinda wasteful to have multiple identical `bindings.rs` files in the project.

Slabity:
Honestly I'm not even sure how much of a difference the 32-bit and 64-bit bindings are themselves. The upstream headers are all quite explicit about what sizes to use and all pointers are captured as u64 regardless of architecture. Even padding appears to be explicit in the structures.

I am curious why the enums are now all of type `libc::c_uint` rather than `u32` as they were previously, but that looks to be a fix in `bindgen` to make enums the size of an integer. I will merge.

Slabity:
Thanks @scrblue for fixing the examples as well, I've also updated the x86_64 bindings.

katyo:
@Slabity The bindings for linux x86_64 and aarch64 is entirely same.

```
git diff --no-index drm-ffi/drm-sys/src/platforms/linux/x86_64/bindings.rs drm-ffi/drm-sys/src/platforms/linux/aarch64/bindings.rs
```

PR #75, opened 2021-03-31
Use drm-fourcc instead of manually maintaining format list

Drakulix:
Proposal:

Instead of manually maintaining a list of drm formats extracted from the various macros of the "drm/drm_fourcc.h" header file,
use [drm-fourcc](https://github.com/danielzfranklin/drm-fourcc-rs), which autogenerates enums out of these macros instead.

- Removes a maintenance burden.
- Updating fourcc with more formats is easier then releasing a new version of drm-rs every time.
- Allows to support more formats and even typed modifiers.
- Allows gbm-rs and smithay to share the same set of formats by also depending on fourcc (I will work on this next).


Slabity:
Good idea. I was not even aware this crate was a thing.

PR #79, opened 2021-04-08
Provide proper property blob get and create functions

Drakulix:
- Fixup `create_property_blob` so that is does not only work with `Mode` but any data.
- Provide `get_property_blob` to receive any data stored in a property blob.

I have opted for allocation here, because the size of property-blob can vary a lot and I could not estimate a reasonable upper bound. Especially EDIDs of connected monitors, which can be read in as a blob, can grow very large depending on the amount of supported modes.
`drm-ffi` provides a perfectly fine `no_std` version already anyway.

### DIFF
--- a/drm-ffi/src/mode.rs
+++ b/drm-ffi/src/mode.rs
@@ -526,7 +526,7 @@ pub fn set_connector_property(
 pub fn get_property_blob(
     fd: RawFd,
     blob_id: u32,
-    data: Option<&mut &mut [u64]>,
+    data: Option<&mut &mut [u8]>,
 ) -> Result<drm_mode_get_blob, Error> {
     let mut blob = drm_mode_get_blob {
         blob_id,
@@ -544,7 +544,7 @@ pub fn get_property_blob(
 }
 
 /// Create a property blob
-pub fn create_property_blob(fd: RawFd, data: &mut [u64]) -> Result<drm_mode_create_blob, Error> {
+pub fn create_property_blob(fd: RawFd, data: &mut [u8]) -> Result<drm_mode_create_blob, Error> {
     let mut blob = drm_mode_create_blob {
         data: data.as_ptr() as _,
         length: data.len() as _,
--- a/examples/atomic_modeset.rs
+++ b/examples/atomic_modeset.rs
@@ -133,7 +133,7 @@ pub fn main() {
         property::Value::CRTC(Some(crtc.handle())),
     );
     let blob = card
-        .create_property_blob(mode)
+        .create_property_blob(&mode)
         .expect("Failed to create blob");
     atomic_req.add_property(
         crtc.handle(),
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -471,18 +471,22 @@ pub trait Device: super::Device {
         Ok(())
     }
 
-    /// Create a property blob value from a given Mode
-    fn create_property_blob(&self, mode: Mode) -> Result<property::Value<'static>, SystemError> {
-        let mut raw_mode: ffi::drm_mode_modeinfo = mode.into();
+    /// Create a property blob value from a given data blob
+    fn create_property_blob<T>(&self, data: &T) -> Result<property::Value<'static>, SystemError> {
         let data = unsafe {
-            std::slice::from_raw_parts_mut(
-                &mut raw_mode as *mut _ as *mut u64,
-                mem::size_of::<ffi::drm_mode_modeinfo>(),
-            )
+            std::slice::from_raw_parts_mut(data as *const _ as *mut u8, mem::size_of::<T>())
         };
         let blob = ffi::mode::create_property_blob(self.as_raw_fd(), data)?;
 
         Ok(property::Value::Blob(blob.blob_id.into()))
+    }
+
+    /// Get a property blob's data
+    fn get_property_blob(&self, blob: u64) -> Result<Vec<u8>, SystemError> {
+        let len = ffi::mode::get_property_blob(self.as_raw_fd(), blob as u32, None)?.length;
+        let mut data = vec![0u8; len as usize];
+        let _ = ffi::mode::get_property_blob(self.as_raw_fd(), blob as u32, Some(&mut &mut *data))?;
+        Ok(data)
     }
 
     /// Destroy a given property blob value


drm-ffi/src/mode.rs
```diff
@@ -536,13 +536,13 @@ pub fn get_property_blob(
         ioctl::mode::get_blob(fd, &mut blob)?;
     }
 
-    map_shrink!(data, blob.length as usize);
+    map_shrink!(&mut data, blob.length as usize);
```
MarijnS95: Is it necessary to update the fat pointer stored in `data: Option`, when the caller doesn't get to see this? It's not passing a _mutable reference_ to the _mutable slice_ (`&mut &mut [u8]`) anymore.

EDIT: Then you can also remove the `mut` again, because no line beyond this point reads the updated value of `data` anyway.

Drakulix: Good point. But the solution here is probably, that this call should instead accept a `&mut &mut` again to make a smaller size visible to the user, without requiring a second call to figure out the length.

MarijnS95: Yup, essentially changing it back to the original type/`fn` signature :). Wasn't sure whether removing `&mut` was actually intended.

Drakulix: Nope, was an unintended mistake. Thanks for noticing!

MarijnS95:
@Drakulix A slightly related question based on the review thread above: why are most of these `drm-ffi` functions returning the raw (unsafe) `drm_*` struct crated from input parameters? In most cases the ioctl returns no extra info through the struct, or the output is provided back through a `&mut &mut [u8]` or similar parameter. If it does, that is possibly better unpacked where possible.

Should we clean up / curate these functions for their return type some time?

Drakulix:
> @Drakulix A slightly related question based on the review thread above: why are most of these `drm-ffi` functions returning the raw (unsafe) `drm_*` struct crated from input parameters? In most cases the ioctl returns no extra info through the struct, or the output is provided back through a `&mut &mut [u8]` or similar parameter. If it does, that is possibly better unpacked where possible.
> 
> Should we clean up / curate these functions for their return type some time?

For the actual motivation of this design, you would have to ask @Slabity.
To me `drm-sys` are raw bindings and `drm-ffi` is a light wrapper on top of the ioctls. And they do partially return data through the info structs (not always of course) and are in that regard *the return type of the ioctl*. (`drm-rs` itself is then the high-level wrapper.)

I am personally in favor of keeping `drm-ffi` consistent in this regard, but I am not a user of that library. If there is a call for a more explicit drm-ffi-api from it's user base, we could likely agree on that kind of change/cleanup. But that should be the subject of a different issue.

MarijnS95:
@Drakulix I see. Tried to make some of these changes myself and the result isn't as interesting as I had hoped. You're right, most likely only `drm-rs` gets to touch these anyway, so it doesn't really matter if there are few cases where none or a single struct member are returning useful data.

It might however be slightly confusing in ie. the FFI example to print a struct containing pointers rather than the appropriately resized slice (interpreted as string might be even better).

Slabity:
`drm-ffi` is suppose to be a safe, minimal wrapper around the commands that `drm-sys` exposes. Unfortunately this kind of puts it in a strange position as some of the commands have a tiny amount of complexity in order to expose a safe interface. Take the `get_bus_id` command for example:

```rust
/// Load this device's Bus ID into a buffer.
///
/// If the buffer is too small, this will load the maximum bytes in the buffer.
/// If the buffer is too big, this will coerce the buffer to the proper size.
pub fn get_bus_id(fd: RawFd, buf: Option<&mut &mut [u8]>) -> Result<drm_unique, Error> {
    let mut busid = drm_unique {
        unique: map_ptr!(&buf),
        unique_len: map_len!(&buf),
    };

    unsafe {
        ioctl::get_bus_id(fd, &mut busid)?;
    }

    map_shrink!(buf, busid.unique_len as usize);

    Ok(busid)
}
```

When `ioctl::get_bus_id` is called, the kernel will look at the `drm_unique` struct passed to it and do two things:

1. Fill up the buffer with `unique_len` amount of data to wherever `unique` is pointing to.
2. Set `unique_len` to the actual size the kernel knows about.

Normally, `libdrm` will call this function twice. First, to find out the size of the bus ID in order to allocate memory for it, and second to fill up the buffer. In order to avoid calling this twice, we pass in an optional `&mut &mut [u8]` value so that not only will the data be filled up, but also the length of the underlying `&mut [u8]` will be adjusted at the same time if necessary.

This means that the returned `drm_unique` struct is only useful when we have the wrong size or pass a `None` to the function in order to get the true size. However, it is certainly useful in that regard.

I 100% agree the design of this is kind of funky, but I'm not sure what a better method would be. And since everything seems to only care about the high-level crate anyway, I never really bothered to think of a better way to handle this.

MarijnS95:
That is exactly the function I was looking at while writing that comment. It seems strange to return the struct when the incoming slice is already mutated to represent the data. For retrieving the length, one could simply return `buslen.unique_len` and not return a raw pointer.

In any case the caller can be heavily simplified regardless of the return type by only looking at the slice reference that is being mutated, going from:

https://github.com/Smithay/drm-rs/blob/cc83dc0753a4a88f337443bf593ecdb809c1b25a/src/lib.rs#L141-L157

To:

```rust
    /// Gets the [BusID] of this device.
    fn get_bus_id(&self) -> Result<BusID, SystemError> {
        let mut buffer = [0u8; 32];
        let mut slice = &mut buffer[..];
        drm_ffi::get_bus_id(self.as_raw_fd(), Some(&mut slice))?;
        Ok(BusID(SmallOsString::from_u8_buffer(buffer, slice.len())))
    }
```

If I were to change this I'd probably make an upfront (stack) allocation with `SmallOsString`, then somehow pass a mutable reference to it that allows updating the length. Something generic like https://doc.rust-lang.org/std/iter/trait.Extend.html might work but that has to move the data again as there is no (unsafe?) function similar to `Vec::set_len`.

> I 100% agree the design of this is kind of funky, but I'm not sure what a better method would be. And since everything seems to only care about the high-level crate anyway, I never really bothered to think of a better way to handle this.

The only alternative coming to mind is two separate functions: one with a `dest: &mut &mut [u8]` argument and returning `()`, the other with no arguments (except `fd`) and returning the size as `i32`.

But I fully agree that it's not worth bothering about this. In hindsight the changes are a bit verbose (though we might want to take advantage of the updated slice length as that saves quite some lines), and it's working as-as anyway. Keeping the high-level bindings in good shape is much more important. @Slabity thanks for your time and reply!

PR #74, opened 2021-03-23
Add CI

katyo:
Solution for #73

Implemented tasks:
- [x] Check code formatting rules
- [x] Check code quality using `clippy`
- [x] Generate bindings for supported platforms (stable Rust)
- [x] Check building on all supported platforms (stable Rust)
- [x] Check building on different Rust channels (stable/beta/nightly, Linux x86_64)
- [x] Run tests on native Linux platforms (x86_64 and i686)
- [x] Create pull-requests for updating generated bindings
- [x] Publish crates when pushing version tag (It requires explicitly setting versions in manifests of crates)

All tasks runs on ubuntu 20.04.
System version of DRM headers is used, so bindings may lack some bleeding edge APIs.
To get publishing working you need set `CRATES_TOKEN` secret.

Currently supported platforms for `bindgen`:
- Linux x86
- Linux x86_64
- Linux arm
- Linux aarch64

Needed for [gbm.rs#4](/Smithay/gbm.rs/pull/4)

PR #81, opened 2021-04-11
Add sub-crates as workspace members; increase CI strictness

MarijnS95:
Noticed in https://github.com/Smithay/drm-rs/pull/74/files#r611252751:

Nightly clippy only runs on modules that are explicitly declared: packages directly in the root `Cargo.toml` and packages declared as workspace members.  `path=` dependencies are treated as normal dependencies (though they don't seem to be on the current 1.51 stable) and are hence not linted.

---

Without `--all-targets` only binaries and libraries are checked or linted. For completeness make sure everything - including the examples - is checked.

---

I can't consistently get `check`/`clippy` results for `build.rs` but managed to fix some warnings when it _did_ eventually run...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all --all-features
+          args: --all --all-features --all-targets
 
   check-minimal:
     runs-on: ubuntu-latest
@@ -117,7 +117,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --all --all-features -Z minimal-versions
+          args: --all --all-features --all-targets -Z minimal-versions
 
   test:
     needs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all --all-features --all-targets -- -D warnings
+          args: --all --all-features --all-targets -- -D warnings -A clippy::redundant_static_lifetimes
 
   check-minimal:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all --all-features --all-targets
+          args: --all --all-features --all-targets -- -D warnings
 
   check-minimal:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,8 @@ jobs:
         with:
           path: target
           key: ${{ runner.os }}-build-rust_nightly-check-${{ hashFiles('**/Cargo.toml') }}
-      - uses: actions-rs/clippy-check@v1
+      - name: Clippy
+        uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all --all-features
@@ -112,7 +113,8 @@ jobs:
         with:
           path: target
           key: ${{ runner.os }}-build-rust_nightly-check-minimal-${{ hashFiles('**/Cargo.toml') }}
-      - uses: actions-rs/cargo@v1
+      - name: Check
+        uses: actions-rs/cargo@v1
         with:
           command: check
           args: --all --all-features -Z minimal-versions
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,9 @@ rustyline = "^8.0.0"
 
 [features]
 use_bindgen = ["drm-ffi/use_bindgen"]
+
+[workspace]
+members = [
+    "drm-ffi",
+    "drm-ffi/drm-sys",
+]
--- a/drm-ffi/src/mode.rs
+++ b/drm-ffi/src/mode.rs
@@ -2,6 +2,8 @@
 //! Bindings to the DRM's modesetting capabilities.
 //!
 
+#![allow(clippy::too_many_arguments)]
+
 use drm_sys::*;
 use ioctl;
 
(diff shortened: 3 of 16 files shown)


MarijnS95:
~This should probably be draft: there are quite a few clippy warnings to be fixed.~ [EDIT: Fixed now!] Additionally I think lint warnings should result in a failed CI (and disallow merge) otherwise there's limited visibility on them and all effort fixing things thus far is lost. This is easily doable by passing `-D warnings` to clippy. @katyo what do you think?

MarijnS95:
Iirc `on: pull_request` triggers run unside the target repository and only pushes to branches within ones fork run inside the fork. Yet, unfortunately:

```
Error: Unable to create clippy annotations! Reason: HttpError: Resource not accessible by integration
Warning: It seems that this Action is executed from the forked repository.
Warning: GitHub Actions are not allowed to create Check annotations, when executed for a forked repos. See https://github.com/actions-rs/clippy-check/issues/2 for details.
```

EDIT: Despite it running in the context of the main repo this makes sense: I could otherwise make a PR with malicious code in the CI script and extract the token that way!

@katyo Perhaps we should not use `actions-rs/clippy-check@v1` until this issue is solved (ie. they stop using the GH token...) and just call `clippy` directly, otherwise this lint has little value until after PRs are merged, triggering CI warnings/errors on master/develop instead.

MarijnS95:
https://github.com/Smithay/drm-rs/runs/2319259289?check_suite_focus=true It's working :)

Unfortunately bindgen generates some code that does not pass this check, we'll likely have to disable `redundant_static_lifetimes` globally for it.

EDIT: The relevant issue is stale for some time already: https://github.com/rust-lang/rust-bindgen/issues/1612

MarijnS95:
Yuck, the latest nightly refuses to compile clang-sys: https://github.com/Smithay/drm-rs/pull/81/checks?check_run_id=2322268251

    nightly-x86_64-unknown-linux-gnu updated - rustc 1.53.0-nightly (a86612456 2021-04-11) (from rustc 1.53.0-nightly (dae9d6ac3 2021-04-09))

drm-ffi/src/mode.rs
```diff
@@ -453,18 +455,18 @@ pub fn set_plane(
...
+        fb_id,
+        flags,
+        crtc_x,
+        crtc_y,
+        crtc_w,
+        crtc_h,
+        src_x,
+        src_y,
+        src_h,
+        src_w,
```
Drakulix, line 469: Did we mess up the order here or did you change this by accident?

MarijnS95: This surprised me as well, but that's actually how it's written in the DRM struct. See https://elixir.bootlin.com/linux/latest/C/ident/drm_mode_set_plane:

```c
struct drm_mode_set_plane {
	__u32 plane_id;
	__u32 crtc_id;
	__u32 fb_id; /* fb object contains surface format type */
	__u32 flags; /* see above flags */

	/* Signed dest location allows it to be partially off screen */
	__s32 crtc_x;
	__s32 crtc_y;
	__u32 crtc_w;
	__u32 crtc_h;

	/* Source values are 16.16 fixed point */
	__u32 src_x;
	__u32 src_y;
	__u32 src_h;
	__u32 src_w;
};
```

`crtc_` is `w/h`, `src_` is `h/w` :shrug:. Not sure what's more readable, following the DRM struct or sensible order? It doesn't matter anyway because Rust requires the field name, but OTOH this very much looks like a C compound literal that _is_ based on order.

.github/workflows/ci.yml
```diff
@@ -84,10 +84,13 @@ jobs:
         with:
           path: target
           key: ${{ runner.os }}-build-rust_nightly-check-${{ hashFiles('**/Cargo.toml') }}
-      - uses: actions-rs/clippy-check@v1
+      - name: Clippy
+        uses: actions-rs/cargo@v1
```
katyo: I think the reasonable alternative is using `actions-rs/clippy-check@v1` conditionally for non-PR pushes.
For ex:
```yaml
    uses: actions-rs/clippy-check@v1
    if: ${{ github.event_name != 'pull_request' }}
    with:
          token: ${{ secrets.GITHUB_TOKEN }}
          args: --all --all-features --all-targets -- -D warnings -A clippy::redundant_static_lifetimes
    uses: actions-rs/cargo@v1
    if: ${{ github.event_name == 'pull_request' }}
    with:
          command: clippy
          args: --all --all-features --all-targets -- -D warnings -A clippy::redundant_static_lifetimes
```

Cargo.toml
```diff
@@ -18,3 +18,9 @@ rustyline = "^8.0.0"
 
 [features]
 use_bindgen = ["drm-ffi/use_bindgen"]
+
+[workspace]
+members = [
+    "drm-ffi",
+    "drm-ffi/drm-sys",
+]
```
katyo, line 26: Sure! Seems I forgot to add this.

MarijnS95: I was surprised to see a bunch of warnings in my editor, and it took a while to notice! `cargo clippy --all` on Rust 1.51 seems to lint these crates regardless of the workspace.

.github/workflows/ci.yml
MarijnS95: I was thinking the same but the suggestion to use a permissions check through yet another Action seemed a bit overkill: https://github.com/actions-rs/clippy-check/issues/2#issuecomment-807878478.

Looking for `PR`s is a simple and elegant solution, let's go with that!

MarijnS95: @katyo I've dropped this change, `clippy-check` totally works without `GITHUB_TOKEN`, it just doesn't create a new task to pretty-print the outputs :tada:! https://github.com/Smithay/drm-rs/runs/2336361898

I probably didn't notice because of missing `-D warnings` and none of the failing crates were in the workspace yet.

MarijnS95: Odd though that GH itself doesn't parse all warnings to show as annotations on the front page, IIRC it used to do that before...

MarijnS95:
@Drakulix @katyo Good news, `nightly` should be fixed and this PR + #82 is now unblocked :grin:

Drakulix:
> This should probably be draft: there are quite a few clippy warnings to be fixed. 

Is this still true? I cannot see any warning in the current (successful) checks for this PR.

MarijnS95:
> > This should probably be draft: there are quite a few clippy warnings to be fixed.
> 
> Is this still true? I cannot see any warning in the current (successful) checks for this PR.

@Drakulix Ah no, they have all been fixed now and the CI is indeed green :)

Drakulix:
thx!

PR #82, opened 2021-04-12
Improve documentation links, wording and fix misspelled words 

MarijnS95:
- Use intra-doc links in more places;
- Always use monospace (``) when referencing items;
- Fix misspelled words.

Might have missed a few, please double-check :)


src/buffer/mod.rs
```diff
@@ -26,7 +26,7 @@
 
 pub use drm_fourcc::{DrmFourcc, DrmModifier, DrmVendor, UnrecognizedFourcc, UnrecognizedVendor};
 
-/// A handle to a GEM buffer.
+/// A handle to a GEM buffer
```
MarijnS95, line 29: Not really sure about line endings. It seems function docs are commonly sentences ended with a period, but type docs are generally left open...

Drakulix:
thx!

PR #88, opened 2021-06-28
drm-ffi: Fix builds on FreeBSD

orvij:
Minor fixes for FreeBSD builds

### DIFF
--- a/drm-ffi/drm-sys/src/platforms/freebsd/x86_64/bindings.rs
+++ b/drm-ffi/drm-sys/src/platforms/freebsd/x86_64/bindings.rs
@@ -343,7 +343,7 @@ pub struct __max_align_t {
 #[test]
 fn bindgen_test_layout___max_align_t() {
     assert_eq!(
-        ::std::mem::size_of::<__max_align_t>(),
+        ::core::mem::size_of::<__max_align_t>(),
         32usize,
         concat!("Size of: ", stringify!(__max_align_t))
     );
@@ -385,12 +385,12 @@ pub union __mbstate_t {
 #[test]
 fn bindgen_test_layout___mbstate_t() {
     assert_eq!(
-        ::std::mem::size_of::<__mbstate_t>(),
+        ::core::mem::size_of::<__mbstate_t>(),
         128usize,
         concat!("Size of: ", stringify!(__mbstate_t))
     );
     assert_eq!(
-        ::std::mem::align_of::<__mbstate_t>(),
+        ::core::mem::align_of::<__mbstate_t>(),
         8usize,
         concat!("Alignment of ", stringify!(__mbstate_t))
     );
@@ -422,7 +422,7 @@ impl Clone for __mbstate_t {
 }
 impl Default for __mbstate_t {
     fn default() -> Self {
-        unsafe { ::std::mem::zeroed() }
+        unsafe { ::core::mem::zeroed() }
     }
 }
 pub type __rman_res_t = __uintmax_t;
@@ -500,7 +500,7 @@ pub struct pthread_barrierattr {
 pub type pthread_barrierattr_t = *mut pthread_barrierattr;
 pub type pthread_spinlock_t = *mut pthread_spinlock;
 pub type pthread_addr_t = *mut libc::c_void;
-pub type pthread_startroutine_t = ::std::option::Option<
+pub type pthread_startroutine_t = ::core::option::Option<
     unsafe extern "C" fn(arg1: *mut libc::c_void) -> *mut libc::c_void,
 >;
 #[repr(C)]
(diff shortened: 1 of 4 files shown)


drm-ffi/drm-sys/src/platforms/freebsd/x86_64/bindings.rs
```diff
@@ -343,7 +343,7 @@ pub struct __max_align_t {
 #[test]
 fn bindgen_test_layout___max_align_t() {
     assert_eq!(
-        ::std::mem::size_of::<__max_align_t>(),
+        ::core::mem::size_of::<__max_align_t>(),
```
orvij, line 346: This diff looks larger than it really is, and basically just consists of `s/std/core` replacements.

There are also the derive impls for `Eq`, `PartialEq`, and `Hash` on three of the `drm_mode_*` structs.

Drakulix:
I am sorry, but I am not sure, that I understand why these changes are necessary for FreeBSD.

Does FreeBSD not have an std?

Also the last two commits do not seem to be related at all.

orvij:
> I am sorry, but I am not sure, that I understand why these changes are necessary for FreeBSD.

I tried building in FreeBSD without these changes, the build fails. The build succeeds with these changes.

Unless something has changed upstream in the last month, these changes are still required.

> Does FreeBSD not have an std?

The bindings are built in a `no_std` environment.

> Also the last two commits do not seem to be related at all.

In what way? They fix breaking builds on FreeBSD.

Drakulix:
> In what way? They fix breaking builds on FreeBSD. Each of the last two commits fix naming ambiguities related to the `ioctl` import, and adding derives for necessary traits. I am pretty descriptive in their purpose in the commit messages.

Sorry if that came out wrong, but I am wondering, why this ambiguity is a problem on FreeBSD builds but not on Linux, that does not really make any sense to me.

Same thing for the derives, they are a purely additional trait implementation, that should cause build failure for linux as well, if it would be required.

So while I have no problems adding those two commits, I am wondering what causes this to avoid such breakage in the future.

orvij:
> Sorry if that came out wrong, but I am wondering, why this ambiguity is a problem on FreeBSD builds but not on Linux, that does not really make any sense to me.

I'm not exactly sure, either, but that is what happens. The ioctl commit fixes it. Maybe there is some conditional compilation that changes `rustc` logic on FreeBSD.

> Same thing for the derives, they are a purely additional trait implementation, that should cause build failure for linux as well, if it would be required.

Not sure about this. I did double-check to make sure this was still the case, and the derives commit is also needed.

Adding a FreeBSD box to the CI build will likely make this necessity obvious.

> So while I have no problems adding those two commits, I am wondering what causes this to avoid such breakage in the future.

I can try to build a minimal reproducer to open an issue in `rust-lang/rust`, but these commits seem the shortest path to support for FreeBSD.

drm-ffi/drm-sys/src/platforms/freebsd/x86_64/bindings.rs
MarijnS95: @orvij This file is automatically generated by `drm-ffi/drm-sys/build.rs`, and a re-run of that will overwrite all your changes. It looks like this file for FreeBSD hasn't been regenerated in four years when `.use_core()` was only added to the bindgen setup a year ago: 6ea69747903a130349c62b116f1faf27e43d3e47. I suggest you run `cargo b --manifest-path drm-ffi/drm-sys/Cargo.toml --features update_bindings` on FreeBSD and commit the result instead. Note that this build script does not always re-run (it should have been a runnable binary, IMO) and might need a `cargo clean` before.

EDIT: The missing `derive` impls were added 3 years ago in https://github.com/Smithay/drm-rs/commit/bb98fe91021ea4486e10660e6208f5d809ca4aa2#diff-7bbd1027edf1cfda4d87b7dcd34ddc1eaafb4984039a88bf72b8a6f5a2950ce4R29-R33, so simply regenerating it should solve all the issues manually covered here.

MarijnS95:
> I'm not exactly sure, either, but that is what happens. The ioctl commit fixes it. Maybe there is some conditional compilation that changes `rustc` logic on FreeBSD.

Any error message you can share? I'm curious too, and that might help understanding _what_ it's conflicting with.

> > Same thing for the derives, they are a purely additional trait implementation, that should cause build failure for linux as well, if it would be required.
> 
> Not sure about this. I did double-check to make sure this was still the case, and the derives commit is also needed.

These were added to the other bindings in https://github.com/Smithay/drm-rs/commit/bb98fe91021ea4486e10660e6208f5d809ca4aa2#diff-7bbd1027edf1cfda4d87b7dcd34ddc1eaafb4984039a88bf72b8a6f5a2950ce4R29-R33, but FreeBSD hasn't been regenerated for a while longer (see my comment above). Most code probably came to rely on these `derive` impls existing.

Drakulix:
> Any error message you can share? I'm curious too, and that might help understanding what it's conflicting with.

Yes please, I do not have a good feeling about this, because right now we are prone to break this again without either *understanding* why this happens or *CI* in place to catch any breakage.

Of course I would merge fixes for any BSD-builds, but please either document why a particular change is necessary or introduce proper CI-testing, because I cannot test this.

PR #93, opened 2021-07-26
drm-ffi/sys: Re-run build.rs when output files change

MarijnS95:
`build.rs` does not re-run unless a clean build is performed.  To easen and solidify bindings regeneration this makes the build script re-run if the output is outdated, or doesn't exist altogether (ie.  intentionally removing the file(s) to force regeneration).  It is also documented in the [first build script example].

Theoretically the input headers (`drm.h`, `drm_mode.h` and everything recursively included) should be printed as well for consistency, but it looks like said feature has not landed in `bindgen` yet.

[first build script example]: https://doc.rust-lang.org/cargo/reference/build-scripts.html


Drakulix:
Looks good! :+1:

PR #94, opened 2021-08-05
Support type flags in Mode

tomstokes:
Naive implementation exposes raw type field as u32 because the Mode
implementation uses the raw ffi object directly

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -1042,6 +1042,11 @@ impl Mode {
     pub fn vrefresh(&self) -> u32 {
         self.mode.vrefresh
     }
+
+    /// Returns the type flags bitmask of this mode
+    pub fn type_(&self) -> u32 {
+        self.mode.type_
+    }
 }
 
 impl From<ffi::drm_mode_modeinfo> for Mode {
@@ -1067,8 +1072,31 @@ impl std::fmt::Debug for Mode {
             .field("hskew", &self.hskew())
             .field("vscan", &self.vscan())
             .field("vrefresh", &self.vrefresh())
+            .field("type_", &self.type_())
             .finish()
     }
+}
+
+/// Display mode type flags
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ModeType {
+    /// Deprecated builtin mode type
+    Builtin = ffi::DRM_MODE_TYPE_BUILTIN,
+    /// Deprecated CLOCK_C mode type
+    ClockC = ffi::DRM_MODE_TYPE_CLOCK_C,
+    /// Deprecated CRTC_C mode type
+    CrtcC = ffi::DRM_MODE_TYPE_CRTC_C,
+    /// Preferred mode
+    Preferred = ffi::DRM_MODE_TYPE_PREFERRED,
+    /// Deprecated default mode
+    Default = ffi::DRM_MODE_TYPE_DEFAULT,
+    /// User defined mode type
+    UserDef = ffi::DRM_MODE_TYPE_USERDEF,
+    /// Mode is created by driver
+    Driver = ffi::DRM_MODE_TYPE_DRIVER,
+    /// Bitmask of all valid (non-deprecated) mode type flags
+    All = ffi::DRM_MODE_TYPE_ALL,
 }
 
 /// Type of a plane


src/control/mod.rs
```diff
@@ -1067,10 +1072,33 @@ impl std::fmt::Debug for Mode {
...
             .field("vrefresh", &self.vrefresh())
+            .field("type_", &self.type_())
             .finish()
     }
 }
 
+/// Display mode type flags
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ModeType {
```
MarijnS95, line 1083: With these being bitflags they should probably be implemented using the `bitflags!` crate as an enum doesn't represent multiple variants at once.

At the same time isn't it easier to expose this publicly in `type_()`? Makes it immediately clear what values (type, really) this function can possibly return, gives easy access to things like `.contains()`, and still allows converting to the underlying raw `u32` with `.bits()`.

Drakulix: drm-rs currently does not depend on `bitflags` and I am not sure we want to.

This style is inline with what we are doing today. Which is [using arrays of flags](https://github.com/Smithay/drm-rs/blob/develop/src/control/mod.rs#L761) *or* [just using integer types](https://github.com/Smithay/drm-rs/blob/develop/src/control/mod.rs#L359).

I do not think we have an example of returning bitflags in the library yet, but the `ModeType` as proposed still has value for testing through e.g. `mode.type_() & (ModeType::Preferred as u32)`.

This is a bit more cumbersome, but I think I am in favor of including it this way for now and discuss the usage of `bitflags` somewhere else. If we would start using that crate, we *should* use it consistently through-out the whole crate, which will then be a breaking change for a `0.6.0` version, while this can be shipped as `0.5.1`.

src/control/mod.rs
```diff
@@ -1067,10 +1072,33 @@ impl std::fmt::Debug for Mode {
...
+    /// Deprecated builtin mode type
+    Builtin = ffi::DRM_MODE_TYPE_BUILTIN,
```
Drakulix, line 1085: Can you add the `#[deprecated]` attribute for deprecated variants?

src/control/mod.rs
```diff
@@ -1067,10 +1072,33 @@ impl std::fmt::Debug for Mode {
             .field("hskew", &self.hskew())
             .field("vscan", &self.vscan())
             .field("vrefresh", &self.vrefresh())
+            .field("type_", &self.type_())
```
Drakulix, line 1075: Can you remove the `_` from the field name string?

src/control/mod.rs
```diff
@@ -1042,6 +1042,11 @@ impl Mode {
     pub fn vrefresh(&self) -> u32 {
         self.mode.vrefresh
     }
+
+    /// Returns the type flags bitmask of this mode
+    pub fn type_(&self) -> u32 {
```
Drakulix, line 1047: Tbh I find the underscore in the name kinda ugly, although this is common practice in rust.
This is a nitpick, but I would prefer something like `mode_type` instead.

Drakulix (changes requested):
Thanks for the contribution!
Some minor nitpicks, but overall this looks very straight-forward! :+1:

src/control/mod.rs
MarijnS95: > drm-rs currently does not depend on `bitflags` and I am not sure we want to.

I certainly hope this can be done at some point, it makes working with bitflags much easier than Rust enums and slices.

> This style is inline with what we are doing today. Which is [using arrays of flags](https://github.com/Smithay/drm-rs/blob/develop/src/control/mod.rs#L761) _or_ [just using integer types](https://github.com/Smithay/drm-rs/blob/develop/src/control/mod.rs#L359).
> 
> I do not think we have an example of returning bitflags in the library yet, but the `ModeType` as proposed still has value for testing through e.g. `mode.type_() & (ModeType::Preferred as u32)`.

Would look something like https://github.com/MarijnS95/drm-rs/commits/bitflags, but not applied to `ModeType` and others just yet. It feels weird to accept integer bitflags - which are supposed to be easily representable - as an array of integers holding the individual bits. I don't think `drm-rs` separates them out anywhere into a vec anyway, and a `u32` doesn't represent the possible return values in the way I would expect from a high-level API.

For this PR the least that could be added is:

```rust
/// Returns the [`ModeType`] bitmask of this mode
```

To make it clear what the possible values are - same for the (linked) uses of `flags: u32` if we don't end up going with bitflags.

> This is a bit more cumbersome, but I think I am in favor of including it this way for now and discuss the usage of `bitflags` somewhere else. If we would start using that crate, we _should_ use it consistently through-out the whole crate, which will then be a breaking change for a `0.6.0` version, while this can be shipped as `0.5.1`.

Fully agreed. It doesn't make sense to introduce that in this PR while nothing else uses it just yet. If you like what you see above I can possibly set up a WIP PR and convert a few more occurrences. Not sure what to do with `DRM_MODE_PAGE_FLIP_TARGET` just yet, though.

Note that quite a few crates in our dependency graph use `bitflags` already.

MarijnS95: Anything still has to happen here? Looking at `DRM_MODE_PAGE_FLIP_TARGET` it seems either `DRM_MODE_PAGE_FLIP_TARGET_ABSOLUTE` or `DRM_MODE_PAGE_FLIP_TARGET_RELATIVE` should be passed in the _`flags`_ field, and the `reserved`/`sequence` field should be used for an integer denoting the desired vblank instead (instead of the flag). We might want to fix that.

See the last commit on aforementioned branch on a potential way to address it. Theoretically `PageFlipTarget` should be merged with `PageFlipFlags`, and we can decide whether to validate it with an incoming value in the `Option<>` as well?

aurabindo:
@Drakulix  Original contributor seems to have abandoned this. I've made the changes you requested here: https://github.com/Smithay/drm-rs/pull/101

Drakulix:
Closed in favor of #101

PR #96, opened 2021-09-29
drm-ffi: Add getfb2 ioctl support

linkmauve:
This ioctl was added in Linux 5.7, and lets a DRM master access a previous framebuffer, no matter its format.  This can be used for smooth handoff between different DRM masters, even with modifiers, multiple planes, etc.

Edit: requires #97 to be merged for CI to complete.

### DIFF
--- a/drm-ffi/src/ioctl.rs
+++ b/drm-ffi/src/ioctl.rs
@@ -125,6 +125,7 @@ pub(crate) mod mode {
 
     /// FB related functions
     ioctl_readwrite!(get_fb, DRM_IOCTL_BASE, 0xAD, drm_mode_fb_cmd);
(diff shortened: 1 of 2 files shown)


Drakulix:
Is there a reason to not use get_fb2, if it is available?
If not, I think we should unify these functions, as they have exactly the same signature.

linkmauve:
They don’t, `get_fb()` doesn’t support modifiers, planes and all of the nice things, while `get_fb2()` does (there is a 2 in the struct name you might have missed).

The only reason not to use the latter is if the user is running a kernel older than 5.7 (or perhaps some BSD).

Drakulix:
Can you rebase this? If the checks are green, I am going to merge this.

linkmauve:
There, rebased.

Drakulix:
Thanks!

PR #97, opened 2021-09-29
Always update apt cache before install

linkmauve:
This prevents CI from failing intermittently.

Edit: requires #98 to be merged for CI to complete.


Drakulix:
Thanks!

PR #101, opened 2021-10-05
Support type flags in Mode

aurabindo:
Naive implementation exposes raw type field as u32 because the Mode
implementation uses the raw ffi object directly

Signed-off-by: Tom Stokes <tomstokes@gmail.com>
Signed-off-by: Aurabindo Pillai <mail@aurabindo.in>

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -1042,6 +1042,11 @@ impl Mode {
     pub fn vrefresh(&self) -> u32 {
         self.mode.vrefresh
     }
+
+    /// Returns the [`ModeType`] flags bitmask of this mode
+    pub fn mode_type(&self) -> u32 {
+        self.mode.flags
+    }
 }
 
 impl From<ffi::drm_mode_modeinfo> for Mode {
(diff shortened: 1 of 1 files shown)


src/control/mod.rs
```diff
@@ -1067,10 +1072,36 @@ impl std::fmt::Debug for Mode {
...
+    /// Deprecated CLOCK_C mode type
+    #[deprecated]
+    ClockC = ffi::DRM_MODE_TYPE_CLOCK_C,
+    /// Deprecated CRTC_C mode type
+    #[deprecated]
+    CrtcC = ffi::DRM_MODE_TYPE_CRTC_C,
+    /// Preferred mode
+    Preferred = ffi::DRM_MODE_TYPE_PREFERRED,
+    /// Deprecated default mode
+    Default = ffi::DRM_MODE_TYPE_DEFAULT,
```
```suggestion
    /// Deprecated default mode
    #[deprecated]
    Default = ffi::DRM_MODE_TYPE_DEFAULT,
```
?

Nit, fwiw: I don't think we need to repeat the word `Deprecated` in the docs, as that's already covered by the `#[deprecated]` annotation:

image

src/control/mod.rs
```diff
@@ -1042,6 +1042,11 @@ impl Mode {
     pub fn vrefresh(&self) -> u32 {
         self.mode.vrefresh
     }
+
+    /// Returns the type flags bitmask of this mode
+    pub fn mode_type(&self) -> u32 {
```
```suggestion
    /// Returns the [`ModeType`] flags bitmask of this mode
    pub fn mode_type(&self) -> u32 {
```

aurabindo:
Updated with the suggested changes.

aurabindo:
Ping @Drakulix  @MarijnS95

MarijnS95 (approved):

Drakulix (approved):
thx!

PR #102, opened 2021-10-18
control: Pass integer sequence for absolute/relative page flips

MarijnS95:
As per the original implementation [1] of the absolute and relative page flipping mechanism:
- The sequence/reserved field should not hold these `DRM_MODE_PAGE_FLIP_TARGET_*` flags but the sequence number;
- The flag is instead passed through the `flags` member;
- Either `ABSOLUTE` or `RELATIVE` is supposed to be passed, not both (which is `DRM_MODE_PAGE_FLIP_TARGET`, and would always result in `-EINVAL`);

[1]: https://git.kernel.org/torvalds/c/f837297ad82480024d3ad08cd84f6670bcafa862

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -759,11 +759,11 @@ pub trait Device: super::Device {
         handle: crtc::Handle,
         framebuffer: framebuffer::Handle,
         flags: &[PageFlipFlags],
-        target: Option<PageFlipTarget>,
+        target_sequence: Option<(PageFlipTarget, u32)>,
     ) -> Result<(), SystemError> {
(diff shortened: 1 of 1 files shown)


MarijnS95:
@Drakulix Anything I can do here and in #103 to receive a review? Would like to see these rather trivial PRs merged so that they can be crossed off my outstanding-PR list :grin:

Drakulix:
> @Drakulix Anything I can do here and in #103 to receive a review? Would like to see these rather trivial PRs merged so that they can be crossed off my outstanding-PR list grin

Sorry, just slipped under my radar. I'll try to get to them later today!

MarijnS95:
Happens to the best of us :) - no worries!

PR #103, opened 2021-10-18
property/control: Support non-linear enum->string conversion

MarijnS95:
On a driver where certain properties not necessarily start at zero, such as an Intel card with the "scaling mode" property omitting the `None` variant and only describing the values 1, 2, and 3, translating a value of 3 ("Full aspect") runs into an index-out-of-bounds panic.  Similarly many properties present an unsorted array of values.

Both problems are solved by initially checking whether the value is out of bounds and otherwise comparing if that value interpreted as index yields the same value from the values array - this is merely a performance optimization and if not, the actual index for that value needs to be linearly scanned before the translation to its string representation can be made.

---

Alternatively a `BTreeMap` could be constructed to simplify the lookup, but it seems drm-rs prefers to postpone the cost of such "expensive" objects/translation to moment it's used instead. It doesn't allow the `values[value] == value` optimization either which is by far the most common enum representation.

### DIFF
--- a/src/control/property.rs
+++ b/src/control/property.rs
@@ -155,7 +155,7 @@ pub enum Value<'a> {
     /// Signed range value
     SignedRange(i64),
     /// Enum Value
-    Enum(&'a EnumValue),
+    Enum(Option<&'a EnumValue>),
     /// Bitmask value
     Bitmask(u64),
     /// Opaque (blob) value
@@ -186,7 +186,7 @@ impl<'a> From<Value<'a>> for RawValue {
             Value::Boolean(false) => 0,
             Value::UnsignedRange(x) => x,
             Value::SignedRange(x) => x as u64,
-            Value::Enum(val) => val.value(),
+            Value::Enum(val) => val.map_or(0, EnumValue::value),
             Value::Bitmask(x) => x,
             Value::Blob(x) => x,
             Value::Object(x) => unsafe { tm::<_, u32>(x).into() },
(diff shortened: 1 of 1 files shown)


src/control/property.rs
```diff
@@ -249,8 +249,14 @@ impl EnumValues {
     ///
     /// Note: This is a dumb translation, not every [`RawValue`] is part of en Enum
```
MarijnS95: This comment is out of date now. Previously the translation would either return the wrong textual value (if the value-as-index is in bounds) or panic if it's out of bounds.

Now the necessary translation is performed, and the function panics if the value is unknown. Alternatively we can return an `Option` or `Result` if that's desired?

Drakulix: I guess, if we are touching this function anyway to properly handle it, an `Option` (along with updating the doc comment here) is a better idea than panicking.

MarijnS95: Great, updated. I've now pretty much just moved the `unwrap` to `convert_value` which isn't desired for the user either. Before writing up a change, what do you prefer here?:

- Turn `Value::Enum(&'a EnumValue)` into `Value::Enum(Option<&'a EnumValue>)`;
- Return an `Option` directly from `convert_value`: this allows us to check the range for `(Un)SignedRange` too, and perhaps a few others, insert null checks etc?
  EDIT: All the reference, ie. `Object`, `CRTC`... are already `Option` - we could factor these out;
  EDIT2: These are also used in `add_property` - does it make sense to pass `None` or a null pointer here?
- Add an extra `Value::Invalid` or just return `Value::Unknown(RawValue)`?

Drakulix: - `Value::Unknown` would be misleading, I guess, because it should be an Enum, we just failed to convert it.
- `Value::Invalid` also sounds weird, because usually the kernel should not send something invalid?
- I think the sensible approach would be just to do `Value::Enum(Option<&'a EnumValue>)`. I agree, that the `Option`-values could be factored out, but then the user would loose the information of the `Value`-type, which previously would have been returned by `convert_value`, but is now just `None`. I am not sure, how valuable this information really is, but if we want to simplify/refactor this code, I would just do that in a separate issue/PR and move on with this.

MarijnS95: Fair enough. At that point we can also have a `Result<>` that could explicitly clarify what the failed-to-parse type is, and if it was ie. out of bounds, null, or not a valid enum value - all seem unlikely to happen.

src/control/property.rs
```diff
@@ -186,7 +186,7 @@ impl<'a> From<Value<'a>> for RawValue {
             Value::Boolean(false) => 0,
             Value::UnsignedRange(x) => x,
             Value::SignedRange(x) => x as u64,
-            Value::Enum(val) => val.value(),
+            Value::Enum(val) => val.map_or(0, EnumValue::value),
```
MarijnS95, line 189: Now again not sure about this one :grimacing:

Drakulix: At this point we would actually need an EnumValue::Unknown to handle the back-and-forth conversion. But I think this is good enough at this point.

Drakulix:
Looks good to me. Anything else or is this good to merge?

MarijnS95:
@Drakulix all good from my side, thanks!

PR #104, opened 2021-10-18
Migrate enum arrays to the bitflags crate

MarijnS95:
Depends on #102

`bitflags` is the "de-facto standard" crate when it comes to representing bitfields in an easy manner, and providing the necessary tooling around it.  It allows to easily construct and compare combinations of flags without having to fall back to casts and slices of individual flag bits.  In addition the type documents and guards possible values, giving a more ergonomic user-experience overall.

This still needs the documentation added in some places.

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -866,7 +866,7 @@ impl Iterator for Events {
             let event = unsafe { &*(self.event_buf.as_ptr().add(self.i) as *const ffi::drm_event) };
             self.i += event.length as usize;
             match event.type_ {
-                x if x == ffi::DRM_EVENT_VBLANK => {
+                ffi::DRM_EVENT_VBLANK => {
                     let vblank_event =
                         unsafe { &*(event as *const _ as *const ffi::drm_event_vblank) };
                     Some(Event::Vblank(VblankEvent {
@@ -878,7 +878,7 @@ impl Iterator for Events {
                         crtc: unsafe { mem::transmute(vblank_event.user_data as u32) },
                     }))
                 }
-                x if x == ffi::DRM_EVENT_FLIP_COMPLETE => {
+                ffi::DRM_EVENT_FLIP_COMPLETE => {
                     let vblank_event =
                         unsafe { &*(event as *const _ as *const ffi::drm_event_vblank) };
                     Some(Event::PageFlip(PageFlipEvent {
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -394,13 +394,13 @@ pub trait Device: super::Device {
             Some(&mut enum_slice),
         )?;
 
+        let flags = ModePropFlags::from_bits_truncate(info.flags);
         let val_len = val_slice.len();
 
         let val_type = {
             use self::property::ValueType;
-            let flags = info.flags;
 
-            if flags & ffi::DRM_MODE_PROP_RANGE != 0 {
+            if flags.contains(ModePropFlags::RANGE) {
                 let min = values[0];
                 let max = values[1];
 
@@ -408,12 +408,12 @@ pub trait Device: super::Device {
                     (0, 1) => ValueType::Boolean,
                     (min, max) => ValueType::UnsignedRange(min, max),
                 }
-            } else if flags & ffi::DRM_MODE_PROP_SIGNED_RANGE != 0 {
+            } else if flags.contains(ModePropFlags::SIGNED_RANGE) {
                 let min = values[0];
                 let max = values[1];
 
                 ValueType::SignedRange(min as i64, max as i64)
-            } else if flags & ffi::DRM_MODE_PROP_ENUM != 0 {
+            } else if flags.contains(ModePropFlags::ENUM) {
                 let enum_values = self::property::EnumValues {
                     values,
                     enums: unsafe { mem::transmute(enums) },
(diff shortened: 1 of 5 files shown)


src/control/mod.rs
```diff
@@ -793,19 +794,26 @@ pub trait Device: super::Device {
...
+    /// Flags to alter the behaviour of a page flip
+    ///
+    /// Note that [`ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET`] bits are
+    /// passed through [`PageFlipTarget`].
+    pub struct PageFlipFlags : u32 {
+        /// Request a vblank event on page flip
+        const EVENT = ffi::drm_sys::DRM_MODE_PAGE_FLIP_EVENT;
+        /// Request page flip as soon as possible, not waiting for vblank
+        const ASYNC = ffi::drm_sys::DRM_MODE_PAGE_FLIP_ASYNC;
+        // TODO: Perhaps include PageFlipTarget values for pretty-printing?
```
MarijnS95: I'm not sure how to best represent `PageFlipTarget` and `PageFlipFlags`. At least providing the flags in the latter allows for pretty-printing (only relevant if these can be read from somewhere...), but we don't really want the user to provide these flags for the sake of making the `ABSOLUTE` and `RELATIVE` variants mutually exclusive - and mandate a sequence number to be passed at the same time.

src/control/mod.rs
```diff
@@ -1137,20 +1145,48 @@ impl PropertyValueSet {
...
-    AllowModeset = ffi::drm_sys::DRM_MODE_ATOMIC_ALLOW_MODESET,
-    /// Generate a page flip event, when the changes are applied.
-    PageFlipEvent = ffi::drm_sys::DRM_MODE_PAGE_FLIP_EVENT,
+bitflags::bitflags! {
+    /// Commit flags for atomic mode setting
+    pub struct AtomicCommitFlags : u32 {
+        /// Generate a page flip event, when the changes are applied.
+        const PAGE_FLIP_EVENT = ffi::drm_sys::DRM_MODE_PAGE_FLIP_EVENT;
+        ///
+        const PAGE_FLIP_ASYNC = ffi::drm_sys::DRM_MODE_PAGE_FLIP_ASYNC;
```
MarijnS95, line 1156: This value was missing but it's included in `DRM_MODE_ATOMIC_FLAGS` (checked by `drm_mode_atomic_ioctl`) so it should be allowed here.

src/control/mod.rs
Drakulix: I would say we should only use `bitflags`, where it represents a bit-field (mutually-exclusive), not when the value is actually an enum (not mutually-exclusive). I agree that pretty-printing is nice, but that is a lot less important here.

src/control/mod.rs
Drakulix: Can you add a doc-comment, please?

Drakulix (approved):
While we want to keep the amount of dependencies for drm-rs low, `bitflags` is a very straight-forward and reasonable addition. I am in favour of merging this, if you want to fix it up and remove the draft status.

src/control/mod.rs
MarijnS95: I think you meant the other way around, bitflags allow concurrent setting whereas an enum is always mutually exclusive.

Regardless, this only becomes a problem when parsing an integer into `PageFlipFlags`, the caller has to be careful to call the variant that allows "unknown" enum values, which are to be parsed separately into `PageFlipTarget`. I don't think that ever (needs) to happen, so this is perfectly fine.

src/control/mod.rs
MarijnS95: I'll take the value from `PageFlipFlags`:

```rust
/// Request page flip as soon as possible, not waiting for vblank
```

And modify it accordingly

```rust
/// Request page flip when the changes are applied, not waiting for vblank
```

MarijnS95: By the way, reading this again, I'd love to have a static assert checking that `AtomicCommitFlags::all() == ffi::drm_sys::DRM_MODE_ATOMIC_FLAGS` :)

src/control/mod.rs
```diff
@@ -445,8 +445,9 @@ pub trait Device: super::Device {
         let property = property::Info {
             handle,
             val_type,
-            mutable: info.flags & ffi::DRM_MODE_PROP_IMMUTABLE == 0,
-            atomic: info.flags & ffi::DRM_MODE_PROP_ATOMIC == 0,
+            mutable: !flags.contains(ModePropFlags::IMMUTABLE),
+            // TODO: this used to be the other way around!
+            atomic: flags.contains(ModePropFlags::ATOMIC),
```
MarijnS95: @Drakulix Forgot to leave a comment on this `TODO`... What's your thought? Fwiw, the kernel documents this property as follows:


```rust
/// the [`Self::ATOMIC`] flag is used to hide properties from userspace that
/// is not aware of atomic properties.  This is mostly to work around
/// older userspace (DDX drivers) that read/write each prop they find,
/// witout being aware that this could be triggering a lengthy modeset.
```

MarijnS95:
> While we want to keep the amount of dependencies for drm-rs low, `bitflags` is a very straight-forward and reasonable addition. I am in favour of merging this, if you want to fix it up and remove the draft status.

That's awesome, thanks! Should be all updated now even with the internal-yet-publicly-exposed `ModePropFlags` documented.

src/control/mod.rs
Drakulix: Yeah, this was probably a copy/paste error from the line above. This seems correct. Please just remove the TODO

MarijnS95: Cool, should be done :)

Drakulix:
Awesome! Thx.
I guess once #103 is done, we want a new drm-rs release with these changes.

PR #109, opened 2021-12-26
Fix mode type flags

tomstokes:
Mode type is in mode.type_, not mode.flags

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -1053,7 +1053,7 @@ impl Mode {
 
     /// Returns the bitmask of this mode
     pub fn mode_type(&self) -> ModeTypeFlags {
-        ModeTypeFlags::from_bits_truncate(self.mode.flags)
+        ModeTypeFlags::from_bits_truncate(self.mode.type_)
     }
 }
 


tomstokes:
Thanks for adding bitflags. Makes the API much easier to work with.

However, the mode type comes from `drm_display_mode.type`, not the `.flags` field. Easy fix.

MarijnS95:
Whoops! Looks like it was already introduced like that in 2467a9e232240200c468b381e691d0f4b36b4de0 though, `ModeType` type with `DRM_MODE_TYPE_*` bitmasks and all.

Should we also add a `bitflags!` type and getter for `DRM_MODE_FLAG_*` then, representing this `.flags` field?

tomstokes:
> Should we also add a `bitflags!` type and getter for `DRM_MODE_FLAG_*` then, representing this `.flags` field?

I could add that, too.

Would you prefer it added to this PR, or should I open a separate PR for the `.flags` getter and types?

MarijnS95:
I'm not the maintainer, but I'd prefer to do that in a separate PR as to unblock this bugfix and get it in ASAP.

@tomstokes It looks like `crates.io` had a little hiccup (also saw that on other repositories), would you mind restarting the CI so that it's green when @Drakulix comes along to merge this?

MarijnS95 (approved):

Drakulix:
> > Should we also add a `bitflags!` type and getter for `DRM_MODE_FLAG_*` then, representing this `.flags` field?
> 
> I could add that, too.
> 
> Would you prefer it added to this PR, or should I open a separate PR for the `.flags` getter and types?

Please do another PR, when you got the time!

Thanks for this

PR #112, opened 2022-03-09
update nix

i509VCB:


Slabity:
Would it be worth re-exporting the `nix` crate from `drm-ffi` and then pull it into `drm-rs`? Along with any other shared crates. That would make it so we only need to update the one file in the future.

i509VCB:
The actual issue appeared to be the `^version` bound which makes 0.23 conflict with the current version used here.

Drakulix:
Anyway, this is an improvement.

PR #113, opened 2022-03-12
Add support for retrieving planar frame buffers

spencercw:
Planar formats (such as compressed formats) cannot be retrieved via the existing method.

Some things to consider for review:
- Unlike `Info`, `PlanerInfo` does not derive `Hash` because `UnrecognizedFourcc` does not derive `Hash`.
- `Info` does not have a function exposing `buffer`; is this intentional?
- `PlanarInfo.buffers` uses `buffer::Handle` instead of `framebuffer::Handle` so it may be passed to `control::buffer_to_prime_fd`.
- The ioctl struct returns four modifier values. I have taken after libdrm which ignores all but the first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ exclude = [".gitignore", ".github"]
 [dependencies]
 bitflags = "1"
 drm-ffi = { path = "drm-ffi", version = "0.2.1" }
-drm-fourcc = "^2.0.0"
+drm-fourcc = "^2.2.0"
 nix = "^0.22.0"
 
 [dev-dependencies]
--- a/drm-ffi/src/result.rs
+++ b/drm-ffi/src/result.rs
@@ -31,6 +31,11 @@ pub enum SystemError {
     /// Permission denied.
     PermissionDenied,
 
+    /// An unknown fourcc code was received.
+    ///
+    /// This likely indicates that the drm-fourcc crate needs updating.
+    UnknownFourcc,
+
     /// Unknown system error.
     Unknown {
         /// Unknown [`nix::errno::Errno`] returned by the system call.
@@ -46,6 +51,7 @@ impl fmt::Display for SystemError {
             SystemError::InvalidArgument => "invalid argument",
             SystemError::InvalidFileType => "invalid file type",
             SystemError::PermissionDenied => "permission denied",
+            SystemError::UnknownFourcc => "unknown fourcc",
             SystemError::Unknown { errno } => {
                 return write!(fmt, "unknown system error: {}", errno)
             }
--- a/src/control/framebuffer.rs
+++ b/src/control/framebuffer.rs
@@ -2,8 +2,10 @@
 //!
 //! Process specific GPU buffers that can be attached to a plane.
 
+use buffer;
 use control;
 use drm_ffi as ffi;
+use drm_fourcc::{DrmFourcc, DrmModifier};
 
 /// A handle to a framebuffer
 #[repr(transparent)]
@@ -46,7 +48,7 @@ pub struct Info {
     pub(crate) pitch: u32,
     pub(crate) bpp: u32,
     pub(crate) depth: u32,
-    pub(crate) buffer: u32,
+    pub(crate) buffer: Option<buffer::Handle>,
 }
 
 impl Info {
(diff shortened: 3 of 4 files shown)


Drakulix:
> Planar formats (such as compressed formats) cannot be retrieved via the existing method.

Thanks for working on this!

>     * Unlike `Info`, `PlanerInfo` does not derive `Hash` because `UnrecognizedFourcc` does not derive `Hash`.
To be honest, I am not super happy with storing a `Result` inside the `PlanarInfo` anyway.
`drm-fourcc` provides a `TryFrom`-implementation to be able to recover from receiving an unknown format, however unlike with modifiers a unknown fourcc format indicates, that the `drm-fourcc` crate needs to be updated to generate from a new libdrm-version.

As such I would honestly prefer `get_planar_framebuffer` to fail in these cases.
Alternatively `PlanarInfo` could store the raw value and attempt the conversion on retrieval.

>     * `Info` does not have a function exposing `buffer`; is this intentional?

Nope. Feel free to add.

>     * `PlanarInfo.buffers` uses `buffer::Handle` instead of `framebuffer::Handle` so it may be passed to `control::buffer_to_prime_fd`.

Note that using `control::buffer_to_prime_fd` and `control::prime_fd_to_buffer` can mess up reference-counting of gbm-buffers. The use of this functions is not encouraged in most cases. Anyway `buffer::Handle` seems to be the correct type here.

>     * The ioctl struct returns four modifier values. I have taken after libdrm which ignores all but the first.

I wonder, if we should do better here. Is it possible for different planes to have different modifiers?

spencercw:
> To be honest, I am not super happy with storing a `Result` inside the `PlanarInfo` anyway. `drm-fourcc` provides a `TryFrom`-implementation to be able to recover from receiving an unknown format, however unlike with modifiers a unknown fourcc format indicates, that the `drm-fourcc` crate needs to be updated to generate from a new libdrm-version.
> 
> As such I would honestly prefer `get_planar_framebuffer` to fail in these cases. Alternatively `PlanarInfo` could store the raw value and attempt the conversion on retrieval.

Yeah I didn't particularly like this either. I think requiring a crate upgrade for new fourcc codes is probably fair. Shall I just add something to the `SystemError` enum (say, `UnknownFourcc`)?

> Nope. Feel free to add.

Will do.

> Note that using `control::buffer_to_prime_fd` and `control::prime_fd_to_buffer` can mess up reference-counting of gbm-buffers. The use of this functions is not encouraged in most cases.

Can you expand on what the problem is here? I'm not particularly familiar with DRM, but I'm looking at using it for screen capture.

> I wonder, if we should do better here. Is it possible for different planes to have different modifiers?

Currently, no. The kernel sets the same modifier for all planes. [1] I'm kind of in two minds as to whether it's a good idea to make this assumption or not. The EGL extension, for example, does not make this assumption and has separate modifier fields for each plane. [2]

[1] https://elixir.bootlin.com/linux/v5.16.14/source/drivers/gpu/drm/drm_framebuffer.c#L567
[2] https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt

Drakulix:
> > To be honest, I am not super happy with storing a `Result` inside the `PlanarInfo` anyway. `drm-fourcc` provides a `TryFrom`-implementation to be able to recover from receiving an unknown format, however unlike with modifiers a unknown fourcc format indicates, that the `drm-fourcc` crate needs to be updated to generate from a new libdrm-version.
> > As such I would honestly prefer `get_planar_framebuffer` to fail in these cases. Alternatively `PlanarInfo` could store the raw value and attempt the conversion on retrieval.
> 
> Yeah I didn't particularly like this either. I think requiring a crate upgrade for new fourcc codes is probably fair. Shall I just add something to the `SystemError` enum (say, `UnknownFourcc`)?

:+1: 

> > Note that using `control::buffer_to_prime_fd` and `control::prime_fd_to_buffer` can mess up reference-counting of gbm-buffers. The use of this functions is not encouraged in most cases.
> 
> Can you expand on what the problem is here? I'm not particularly familiar with DRM, but I'm looking at using it for screen capture.

See https://github.com/swaywm/wlroots/pull/2672 and https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/5442.

Re-reading this issue, this is likely not a problem for screen-recording and there is no other way to get a dmabuf from the handle, as you cannot import a raw-handle into libgbm.

> > I wonder, if we should do better here. Is it possible for different planes to have different modifiers?
> 
> Currently, no. The kernel sets the same modifier for all planes. [1] I'm kind of in two minds as to whether it's a good idea to make this assumption or not. The EGL extension, for example, does not make this assumption and has separate modifier fields for each plane. [2]
> 
> [1] https://elixir.bootlin.com/linux/v5.16.14/source/drivers/gpu/drm/drm_framebuffer.c#L567 [2] https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt

In that case we should probably feature-proof this and support multiple modifiers. It is easy enough for downstream to discard the rest.

spencercw:
> Re-reading this issue, this is likely not a problem for screen-recording and there is no other way to get a dmabuf from the handle, as you cannot import a raw-handle into libgbm.

Think I'll just have to take your word on this one. Don't think I've got enough background knowledge to make sense of what's going on here.

Anyway, I've made those changes.

Drakulix (approved):

spencercw:
Updated drm-fourcc to fix build issue with `Hash` derivation.

PR #114, opened 2022-03-30
Add support for waiting for vblanks

spencercw:
- The Linux and FreeBSD bindings for the vblank flags are slightly different (although the values are the same), so I have just defined them manually. I did try using `cfg` to define them conditionally based upon the target OS, but it was getting a bit messy.
- The signal and flip flags don't do anything useful so I have omitted these, although curiously the signal number is copied into vblank event as the user data, so I have put that in the API.
- The secondary flag is equivalent to setting `high_crtc=1`, so I have omitted this.

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -786,19 +786,28 @@ pub trait Device: super::Device {
         handle: crtc::Handle,
         framebuffer: framebuffer::Handle,
         flags: PageFlipFlags,
-        target_sequence: Option<(PageFlipTarget, u32)>,
+        target_sequence: Option<PageFlipTarget>,
     ) -> Result<(), SystemError> {
         let mut flags = flags.bits();
-        if let Some((target, _)) = target_sequence {
-            flags |= target as u32;
-        }
+
+        let sequence = match target_sequence {
+            Some(PageFlipTarget::Absolute(n)) => {
+                flags |= ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET_ABSOLUTE;
+                n
+            }
+            Some(PageFlipTarget::Relative(n)) => {
+                flags |= ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET_RELATIVE;
+                n
+            }
+            None => 0,
+        };
 
         let _info = ffi::mode::page_flip(
             self.as_raw_fd(),
             handle.into(),
             framebuffer.into(),
             flags,
-            target_sequence.map(|(_, sequence)| sequence).unwrap_or(0),
+            sequence,
         )?;
 
         Ok(())
@@ -834,18 +843,17 @@ bitflags::bitflags! {
     }
 }
 
-#[repr(u32)]
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 /// Target to alter the sequence of page flips
 ///
-/// These are the [`ffi::drm_sys::DRM_MODE_PAGE_FLIP_FLAGS`] bits
+/// These represent the [`ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET`] bits
 /// of [`PageFlipFlags`] wrapped in a regular `enum` due to their
 /// mutual-exclusiveness.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum PageFlipTarget {
     /// Absolute Vblank Sequence
-    Absolute = ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET_ABSOLUTE,
+    Absolute(u32),
     /// Relative Vblank Sequence (to the current, when calling)
-    Relative = ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET_RELATIVE,
+    Relative(u32),
 }
 
 /// Iterator over [`Event`]s of a device. Create via [`Device::receive_events()`].
@@ -869,10 +877,12 @@ pub enum Event {
 pub struct VblankEvent {
     /// sequence of the frame
     pub frame: u32,
-    /// duration between events
-    pub duration: Duration,
+    /// time at which the vblank occurred
+    pub time: Duration,
     /// crtc that did throw the event
     pub crtc: crtc::Handle,
+    /// user data that was passed to wait_vblank
+    pub user_data: usize,
 }
 
 /// Page Flip event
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -305,3 +337,42 @@ pub enum ClientCapability {
...
 }
+
+/// Used to specify a vblank frame to wait for
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum VblankWaitFrame {
+    /// Wait for a specific vblank sequence number
+    Absolute(u32),
+    /// Wait for a given number of vblanks
+    Relative(u32),
+}
```
MarijnS95: If you think this is a nice API, it could perhaps also be employed in `fn page_flip` in the `control` module, which currently uses a `(PageFlipTarget, u32)` type for this (alternatively use that format with bitflag enum here?).

src/lib.rs
```diff
@@ -209,6 +210,37 @@ pub trait Device: AsRawFd {
...
+        if high_crtc > 0x1f {
+            return Err(SystemError::InvalidArgument);
+        }
+
+        let (sequence, wait_type) = match frame {
+            VblankWaitFrame::Absolute(n) => (n, drm_ffi::DRM_VBLANK_ABSOLUTE),
+            VblankWaitFrame::Relative(n) => (n, drm_ffi::DRM_VBLANK_RELATIVE),
+        };
+
+        let type_ = wait_type | (high_crtc << 1) | flags.bits();
```
MarijnS95: The kernel still defines `_DRM_VBLANK_HIGH_CRTC_SHIFT`, this just needs a minor tweak to the bindgen build script.

MarijnS95:
> * The Linux and FreeBSD bindings for the vblank flags are slightly different (although the values are the same), so I have just defined them manually. I did try using `cfg` to define them conditionally based upon the target OS, but it was getting a bit messy.

This is only caused by the file not being regenerated on a FreeBSD system in a long time. https://github.com/Smithay/drm-rs/pull/88 could have addressed this but the author disappeared (`drm-rs` is the last activity on their GH profile :disappointed:).

src/lib.rs
```diff
@@ -209,6 +210,37 @@ pub trait Device: AsRawFd {
...
+
+    /// Waits for a vblank.
+    fn wait_vblank(
+        &self,
+        frame: VblankWaitFrame,
+        flags: VblankWaitFlags,
+        high_crtc: u32,
+        user_data: usize,
+    ) -> Result<VblankWaitReply, SystemError> {
+        if high_crtc > 0x1f {
```
MarijnS95: Can the `_DRM_VBLANK_HIGH_CRTC_MASK` constant be used here, after shifted right by `_DRM_VBLANK_HIGH_CRTC_SHIFT`?

MarijnS95:
Fwiw if anyone is running on FreeBSD or has a cross-compiler setup for FreeBSD, running something along the lines of `cargo b --manifest-path drm-ffi/drm-sys/Cargo.toml --features update_bindings --target=x86_64-unknown-freebsd` and contributing the result would be great :+1:

spencercw:
Thanks. I have rebased on your branch and addressed your comments.

src/control/mod.rs
```diff
@@ -834,18 +843,17 @@ bitflags::bitflags! {
     }
 }
 
-#[repr(u32)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 /// Target to alter the sequence of page flips
 ///
-/// These are the [`ffi::drm_sys::DRM_MODE_PAGE_FLIP_FLAGS`] bits
+/// These represent the [`ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET`] bits
```
MarijnS95, line 848: Nice catch ^^ (not sure why it was written correctly in the docs for `PageFlipFlags` above) :+1:

src/control/mod.rs
```diff
@@ -834,18 +843,17 @@ bitflags::bitflags! {
...
 /// of [`PageFlipFlags`] wrapped in a regular `enum` due to their
 /// mutual-exclusiveness.
 pub enum PageFlipTarget {
     /// Absolute Vblank Sequence
-    Absolute = ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET_ABSOLUTE,
+    Absolute(u32),
     /// Relative Vblank Sequence (to the current, when calling)
-    Relative = ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET_RELATIVE,
+    Relative(u32),
 }
```
MarijnS95: This type probably shouldn't be unified with `VblankWaitTarget`, if only for keeping the documentation neatly separated?

Fwiw I agree with you placing doc comments above `#[derive]` attributes, feel free to correct the same here.

src/lib.rs
```diff
@@ -209,6 +210,44 @@ pub trait Device: AsRawFd {
...
+        flags: VblankWaitFlags,
+        high_crtc: u32,
+        user_data: usize,
+    ) -> Result<VblankWaitReply, SystemError> {
+        let high_crtc_shift = drm_ffi::_DRM_VBLANK_HIGH_CRTC_SHIFT;
+        let high_crtc_mask =
+            drm_ffi::drm_vblank_seq_type::_DRM_VBLANK_HIGH_CRTC_MASK >> high_crtc_shift;
+        if (high_crtc & !high_crtc_mask) != 0 {
+            return Err(SystemError::InvalidArgument);
+        }
```
MarijnS95: Nit: could you save on some code/lines by pre-shifting `high_crtc << drm_ffi::_DRM_VBLANK_HIGH_CRTC_SHIFT` and:

1. Using that result below directly;
2. Use the mask against it directly, instead of also shifting the mask (which should be a constant operation, I hope the compiler is able to understand that for `high_crtc_shift` being a `let` binding instead of a `const` binding, which I don't think is needed for readability FWIW).

src/control/mod.rs
spencercw: I did consider merging these, but I think it might be worth keeping them separate, even if only for clarity of documentation. I could go either way though.

MarijnS95: Let's keep it for the sake of clear documentation :+1:

src/lib.rs
spencercw: The double shift was actually intentional because if we shift the parameter before checking it then `1 << 31` is accepted as a valid input.

MarijnS95: That's a real outlier, but fair enough :smile:

MarijnS95: In any case, not sure if the `let high_crtc_shift` really helps with readability compared to a `const` or even importing it in scope.

spencercw: Do you mean something like this?
```rust
fn wait_vblank(...) -> ... {
    use drm_ffi::drm_vblank_seq_type::_DRM_VBLANK_HIGH_CRTC_MASK as HIGH_CRTC_MASK;
    use drm_ffi::_DRM_VBLANK_HIGH_CRTC_SHIFT as HIGH_CRTC_SHIFT;
    
    let high_crtc_mask = HIGH_CRTC_MASK >> HIGH_CRTC_SHIFT;
    if (high_crtc & !high_crtc_mask) != 0 {
        return Err(SystemError::InvalidArgument);
    }
```

MarijnS95: Yeah, that way they stay const without having to do `const THE_SHORTER_NAME: the_type = drm_ffi::..;`, though not sure if you need the `as` rename anyway but could just use the full name without the `drm_ffi::` in the rest of the code.

EDIT: I guess `let high_crtc_mask = HIGH_CRTC_MASK >> HIGH_CRTC_SHIFT;` could become `const` too... Really nitpicking here :)

spencercw: Fair point. Changes pushed.

Drakulix (approved):
Thanks @spencercw for working on this!
Also thanks to @MarijnS95 for doing all the review work beforehand, I have nothing left to complain about. :)

PR #115, opened 2022-03-30
drm-ffi/drm-sys: Generate const for defines starting with `_` before `DRM`

MarijnS95:
CC @spencercw, this should help with #114

~EDIT: Waiting for the CI to send a PR to update bindings for other platforms.~


MarijnS95:
Updated binding for all architectures by overwriting with `bindings.zip` artifact from the CI, without `DRM_MODE_CONNECTOR_USB` from newer headers to match what's used in the CI (Ubuntu focal repos).

PR #118, opened 2022-06-19
fix: unsound transmutes

### LINKED ISSUE #117: transmuting Vec might lead to undefined behavior
Hello, I've come across these transmutes when browsing the code:

https://github.com/Smithay/drm-rs/blob/e5f5b14d58c0d598cbce4019c706b2b1e0cf7d7e/src/control/mod.rs#L170

https://github.com/Smithay/drm-rs/blob/e5f5b14d58c0d598cbce4019c706b2b1e0cf7d7e/src/control/mod.rs#L539

They are transmuting `Vec<drm_mode_modeinfo>` into `Vec<Mode>`. However, `Vec` has an unspecified layout and this can lead to UB. This is even mentioned in the documentation https://doc.rust-lang.org/std/mem/fn.transmute.html#alternatives (scroll to "Turning a `Vec<&T>` into a `Vec<Option<&T>>`.")

The documentation also offers an alternative:
```rs
let after = unsafe {
    let mut before = std::mem::ManuallyDrop::new(before);
    Vec::from_raw_parts(before.as_mut_ptr() as *mut _, before.len(), before.capacity())
};
```

Secondly, I have also notices these two `Handle` types aren't `repr(transparent)` as they should be. Other `Handle` types are, so I assume this is just an oversight. 

https://github.com/Smithay/drm-rs/blob/e5f5b14d58c0d598cbce4019c706b2b1e0cf7d7e/src/control/encoder.rs#L11

https://github.com/Smithay/drm-rs/blob/e5f5b14d58c0d598cbce4019c706b2b1e0cf7d7e/src/control/crtc.rs#L20

I can open a PR to fix these, should be very simple.

TheEdda:
* missing repr(transparent) on some handles
* transmute of Vec which has unspecified layout

closes #117

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -72,12 +72,8 @@ pub fn from_u32<T: ResourceHandle>(raw: u32) -> Option<T> {
 
 unsafe fn transmute_vec<T, U>(from: Vec<T>) -> Vec<U> {
     let mut from = std::mem::ManuallyDrop::new(from);
-    
-    Vec::from_raw_parts(
-        from.as_mut_ptr() as *mut U,
-        from.len(),
-        from.capacity()
-    )
+
+    Vec::from_raw_parts(from.as_mut_ptr() as *mut U, from.len(), from.capacity())
 }
 
 /// This trait should be implemented by any object that acts as a DRM device and
@@ -572,7 +568,7 @@ pub trait Device: super::Device {
 
         let prop_val_set = PropertyValueSet {
             prop_ids: unsafe { mem::transmute(prop_ids) },
-            prop_vals: unsafe { mem::transmute(prop_vals) },
+            prop_vals,
             len: prop_len,
         };
 
@@ -812,7 +808,7 @@ pub trait Device: super::Device {
             None => 0,
         };
 
-        let _info = ffi::mode::page_flip(
+        ffi::mode::page_flip(
             self.as_raw_fd(),
             handle.into(),
             framebuffer.into(),
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -177,7 +177,7 @@ pub trait Device: super::Device {
                 (0, 0) => None,
                 (x, y) => Some((x, y)),
             },
-            modes: unsafe {  transmute_vec(modes) },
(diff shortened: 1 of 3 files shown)


src/control/mod.rs
```diff
@@ -70,6 +70,16 @@ pub fn from_u32<T: ResourceHandle>(raw: u32) -> Option<T> {
     RawResourceHandle::new(raw).map(T::from)
 }
 
+unsafe fn transmute_vec<T, U>(from: Vec<T>) -> Vec<U> {
```
MarijnS95, line 73: Isn't it more common to use the `safe_transmute` crate for these?

src/control/mod.rs
```diff
@@ -167,7 +177,7 @@ pub trait Device: super::Device {
                 (0, 0) => None,
                 (x, y) => Some((x, y)),
             },
-            modes: unsafe { mem::transmute(modes) },
+            modes: unsafe {  transmute_vec(modes) },
```
```suggestion
            modes: unsafe { transmute_vec(modes) },
```

src/control/mod.rs
TheEdda: I suppose that would warrant converting the whole crate to use such a crate (and/or [getting rid of the transmutes in the first place...](https://github.com/Smithay/drm-rs/issues/117#issuecomment-1159784840)). I assumed since this is very limited in scope and has an explicit example in the std docs I would just implement it right there.

Indeed, it is the same code as some other crates provide, excluding size and alignment checks, which might be a benefit of an existing crate (or it can be added to this function):
* https://docs.rs/transvec/latest/src/transvec/lib.rs.html#511
* https://docs.rs/safe-transmute/latest/src/safe_transmute/full.rs.html#298
* https://docs.rs/bytemuck/latest/src/bytemuck/allocation.rs.html#182

Personally I'm partial to bytemuck as it is most popular, but also the heaviest :woman_shrugging:

Drakulix: I think that is out of the scope of this PR. Fixing the currently unsound behaviour should be done first, we can improve upon that later. Feel free to open another issue for discussing the different approaches.

Drakulix (approved):
LGTM! Thanks

PR #119, opened 2022-06-28
Feature/add bytemuck

TheEdda:
based on top of #118 

relevant to #117

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ exclude = [".gitignore", ".github"]
 
 [dependencies]
 bitflags = "1"
+bytemuck = { version = "1.11", features = ["extern_crate_alloc", "derive"] }
 drm-ffi = { path = "drm-ffi", version = "0.2.1" }
 drm-fourcc = "^2.2.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ exclude = [".gitignore", ".github"]
 
 [dependencies]
 bitflags = "1"
-bytemuck = { version = "1.11", features = ["extern_crate_alloc", "derive"] }
+bytemuck = { version = "1.12", features = ["extern_crate_alloc", "derive"] }
 drm-ffi = { path = "drm-ffi", version = "0.2.1" }
 drm-fourcc = "^2.2.0"
 
--- a/examples/kms_interactive.rs
+++ b/examples/kms_interactive.rs
@@ -4,6 +4,7 @@ extern crate rustyline;
 
 /// Check the `util` module to see how the `Card` structure is implemented.
 pub mod utils;
+use drm::control::{from_u32, RawResourceHandle};
 use utils::*;
 
 pub fn main() {
@@ -78,8 +79,7 @@ fn run_repl(card: &Card) {
             // Destroying a framebuffer
             ["DestroyFramebuffer", handle] => {
                 let handle: u32 = str::parse(handle).unwrap();
-                let handle: drm::control::framebuffer::Handle =
-                    unsafe { std::mem::transmute(handle) };
+                let handle: drm::control::framebuffer::Handle = from_u32(handle).unwrap();
                 if let Err(err) = card.destroy_framebuffer(handle) {
                     println!("Unable to destroy framebuffer ({:?}): {}", handle, err);
                 }
@@ -97,7 +97,7 @@ fn run_repl(card: &Card) {
             // Print out the values of a specific property
             ["GetProperty", handle] => {
                 let handle: u32 = str::parse(handle).unwrap();
-                let handle: drm::control::property::Handle = unsafe { std::mem::transmute(handle) };
+                let handle: drm::control::property::Handle = from_u32(handle).unwrap();
                 let property = card.get_property(handle).unwrap();
                 println!("\tName: {:?}", property.name());
                 println!("\tMutable: {:?}", property.mutable());
@@ -124,8 +124,7 @@ fn run_repl(card: &Card) {
             // Set a property's value on a resource
             ["SetProperty", handle, property, value] => {
                 let property: u32 = str::parse(property).unwrap();
-                let property: drm::control::property::Handle =
-                    unsafe { std::mem::transmute(property) };
+                let property: drm::control::property::Handle = from_u32(property).unwrap();
                 let value: u64 = str::parse(value).unwrap();
 
                 match HandleWithProperties::from_str(card, handle) {
@@ -186,7 +185,7 @@ impl HandleWithProperties {
     // the corresponding resource.
     fn from_str(card: &Card, handle: &str) -> Result<Self, ()> {
         let handle: u32 = str::parse(handle).unwrap();
-        let handle = drm::control::RawResourceHandle::new(handle).unwrap();
+        let handle = RawResourceHandle::new(handle).unwrap();
 
         let rhandles = card.resource_handles().unwrap();
         for connector in rhandles.connectors().iter().map(|h| (*h).into()) {
--- a/src/buffer/mod.rs
+++ b/src/buffer/mod.rs
@@ -24,6 +24,7 @@
 //! like a regular one. This allows better control and security, and is the
 //! recommended method of sharing buffers.
 
+use control;
 pub use drm_fourcc::{DrmFourcc, DrmModifier, DrmVendor, UnrecognizedFourcc, UnrecognizedVendor};
 
 /// A handle to a GEM buffer
@@ -35,9 +36,13 @@ pub use drm_fourcc::{DrmFourcc, DrmModifier, DrmVendor, UnrecognizedFourcc, Unre
 /// prevent buffers from leaking by properly closing them after they are done.
 #[repr(transparent)]
 #[derive(Copy, Clone, Hash, PartialEq, Eq)]
-pub struct Handle(::control::RawResourceHandle);
+pub struct Handle(control::RawResourceHandle);
 
-impl From<Handle> for ::control::RawResourceHandle {
+// Safety: Handle is repr(transparent) over NonZeroU32
+unsafe impl bytemuck::ZeroableInOption for Handle {}
+unsafe impl bytemuck::PodInOption for Handle {}
+
+impl From<Handle> for control::RawResourceHandle {
     fn from(handle: Handle) -> Self {
         handle.0
     }
@@ -49,8 +54,8 @@ impl From<Handle> for u32 {
     }
 }
 
-impl From<::control::RawResourceHandle> for Handle {
-    fn from(handle: ::control::RawResourceHandle) -> Self {
+impl From<control::RawResourceHandle> for Handle {
+    fn from(handle: control::RawResourceHandle) -> Self {
         Handle(handle)
     }
 }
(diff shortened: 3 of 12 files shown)


Drakulix:
I want @Slabity 's opinion on this one.

I think it getting rid of the undefined behaviour should be our goal in any case.
Using a crate for that, so we don't have to keep up with best practices ourselves also seems like a reasonable decision.
`bytemuck` is also has practically no dependencies on it's own and has a small footprint.

But it is a bigger change nevertheless, so anyone feel free to raise any concerns rather sooner than later.

Slabity:
So I do actually have a few concerns on this:

1. I'm not quite sure I understand the motivation of using `bytemuck` just for the sake of removing `mem::transmute`. Either the calls to `bytemuck` are doing additional/hidden checks underneath, or it's going to end up being turned into a transmute/optimized out by the compiler anyways. From briefly looking at the internal casting functions of the `bytemuck` crate, it actually looks like it does *both*, which IMHO is kind of the worst of both worlds.
2. I have no idea what the `bytemuck::ZeroableInOption` and `bytemuck::PodInOption` traits are that are being included in various locations or why they are necessary. I can't find any documentation for these in the bytemuck crate at all or understand it from the name alone, which may make it more difficult for future contributors to grok their way around what the code might be doing.
3. If the only concern is transmuting slices due to the (very unlikely) possibility that the slices will have a different layout, then I think it would be simpler to create a function that just breaks down the slice into its length and pointer, and then reconstructs it as as a new slice. I'd expect this to be optimized out by the compiler in cases where the layouts are the same anyways.

I'm not entirely unwilling to use `bytemuck`, but I think I need a more detailed motivation on why it's necessary over the current `core` functions that rust provides (speaking of, we should really look into removing any `std` functions that can be replaced by `core` functions).

I do think the other parts of this MR are good ideas. There's a lot of transmuting different objects and handles where we already have the ability to create them without transmutes anyways. Same with adding missing `#[repr(transparent)]` attributes.

TheEdda:
@Slabity 

1. The `bytemuck` crate is ideally exactly the place where to do all your transmutes. In my eyes (and hopefully in eyes of all its contributors and users, of which there are many) it provides good abstraction over what is safe to transmute and what is just "very unlikely" to be UB. The checks it does are going to get optimized away if the compiler can prove that they will hold (which is exactly when it would optimize a slice cast away too), so I don't belive it's in any way worse than just doing a "manual" transmute **and** it directly provides us with experience of its contributors on this pretty serious issue - case in point, the `transmute_vec` function I introduced here in my previous PR here turned out to have a soudness issue which I didn't know existed, but the [contributors noticed](https://github.com/Lokathor/bytemuck/pull/111#discussion_r906842782) and it was fixed in my PR there.

2. These traits were added by me in https://github.com/Lokathor/bytemuck/pull/114 - they are not yet released and as you can see the Cargo.toml here still links to my forked git version of bytemuck, hence the WIP status of this PR. These traits act _exactly_ as two well-known bytemuck traits [`Zeroable`](https://docs.rs/bytemuck/latest/bytemuck/trait.Zeroable.html) and [`Pod`](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html) but for `Option<T>` where `T` is this crate's custom type `#[repr(transparent)] Handle(NonZeroU32)`. Due to orphan rules it was not possible to implement `Zeroable` and `Pod` on `Option<Handle>` within a downstream crate so I figured it would be a good addition. This crate has a lot of `Option<Handle>` instances which can be safely transmuted from and to `u32` (this is a Rust layouting guarantee). Also it took me some figuring out what the transmute code was doing before - and in some cases my worries about it being potentially "just" layouting UB and in one case even full segfault UB (`kms_interactive` example, it was transmuting from unsanitized user input `u32` into `Handle(NonZeroU32)`) were correct, which I didn't like, hence the issues and PRs to fix this.

3. There are no slices being transmuted anywhere in this crate, not sure where you are getting this. Either way, bytemuck already does this, properly, with correct constrains on the unsafe traits _and_ the additional eyes on the code have more time to track potential changes or sources of soundness bugs over time. I can't stress this enough, if I'm going to mass-transmute I'd rather use a crate that helps me do so safely and has had someone point out most (if not all) the pitfalls I could forget about in the safety contract that just throw some in there and hope for the best.

Anyway, back to this crate - the other issue here is that there are `array`s that need to be transformed. Arrays before const generics were a pain. Const generics (and `array::map`) came in `1.55` which was deemed by @Drakulix as too high to require (eventually, I know there are no checks for MSRV currently and I think I saw some `array::map` in unrelated code in this crate). Since const generics are out, there are two ways to handle this - macros and transmutes. Macros dealing with arrays are always a pain, plus the precedence that there were already transmutes make me believe that safe-transmuting is the way to go. I do hope that eventually the MSRV will rise to/beyond 1.55 and it will be possible to replace them with `array::map`, but in the meantime this is imo the safest available way around this - plus, it still improves safety because it is not possible to  cast `u32 -> Handle`, it needs to be `u32 -> Option<Handle> -> Option::unwrap` which prevents the c-ffi from causing real UB by returning the wrong `u32` (as is tradition).

TheEdda:
Oh god it looks like `syn` broke semver at some point and it's getting caught by the `check-minimal` step :woman_facepalming: (issue https://github.com/dtolnay/syn/issues/1194)

Slabity:
>     2. These traits were added by me in [feat: ZeroableInOption and PodInOption traits Lokathor/bytemuck#114](https://github.com/Lokathor/bytemuck/pull/114) - they are not yet released and as you can see the Cargo.toml here still links to my forked git version of bytemuck, hence the WIP status of this PR. These traits act _exactly_ as two well-known bytemuck traits [`Zeroable`](https://docs.rs/bytemuck/latest/bytemuck/trait.Zeroable.html) and [`Pod`](https://docs.rs/bytemuck/latest/bytemuck/trait.Pod.html) but for `Option<T>` where `T` is this crate's custom type `#[repr(transparent)] Handle(NonZeroU32)`. Due to orphan rules it was not possible to implement `Zeroable` and `Pod` on `Option<Handle>` within a downstream crate so I figured it would be a good addition. This crate has a lot of `Option<Handle>` instances which can be safely transmuted from and to `u32` (this is a Rust layouting guarantee). Also it took me some figuring out what the transmute code was doing before - and in some cases my worries about it being potentially "just" layouting UB and in one case even full segfault UB (`kms_interactive` example, it was transmuting from unsanitized user input `u32` into `Handle(NonZeroU32)`) were correct, which I didn't like, hence the issues and PRs to fix this.

Ah, thanks for the explanation. That makes sense.

About the segfault in `kms_interactive` (which is definitely a mess, I made it quite haphazardly in a few hours to manually test and document how DRM worked on the kernel side) - Do you know what was causing the segfault? Was it from passing a zero into it? And was the segfault when the `ioctl` function was called? I'd like to know if this was a memory segfault or an artificial kernel-initiated segfault based on bad input.

>     3. There are no slices being transmuted anywhere in this crate, not sure where you are getting this..

Meant arrays here. There's a few locations where we are transmuting property arrays or something else.

Honestly I think the underlying issue here is that I somewhat over-engineered the original implementation by referencing the Linux and FreeBSD's DRM subsystem code to try and make it more difficult for users to make a mistake and giving it invalid input. Choosing `NonZeroU32` for various different handles and property representations is something that I think will come back to bite us.

Technically speaking, it is possible for a different kernel to have their own implementation that *doesn't* say zero is an invalid handle or property in certain locations. I forget what the result is for an invalid handle (I believe it's just `EINVAL`, but if it caused the segfault you saw in `kms_interactive` then it will just be something we have to live with). If it is possible to get rid of casting altogether in this case, I would prefer it over to changing the method of casting to begin with.

Unfortunately I also did the same with the array lengths. I manually checked the maximum lengths of arrays the kernel would provide and set those accordingly. I did this to avoid any sort of allocations like `Vec` does, but we ended up adding a `Vec` anyways due to the number of `Modes` being provided exceeding what was originally used in newer kernel versions. This was all before custom allocators became a thing, so I'm much less bothered about allocations now.

Anyways, sorry for the somewhat off-topic banter. It looks like we are already compiling `bytemuck` in our dependencies, so I don't see any technical reason not to include it. However, I don't want to remove any sort of conversion functions that we currently have (such as `from_u32`), and I'd like to use the regular conversion, wrapping, and unwrapping functions wherever possible rather than relying on `transmute` or `cast` of any kind.

TheEdda:
I have rebased on develop and added the `from_u32` function back - although with a slightly different trait bound if that is okay. 

We'll see how the syn version issue goes and if I can get it into bytemuck so that the check-minimal CI starts passing. Then hopefully a bytemuck release with the new traits.

Also has the CI changed? It seems like clippy is complaining about things that were already there...

Drakulix:
> Also has the CI changed? It seems like clippy is complaining about thing that were already there...

A new rust version was released and with it some new clippy lints.
Fixing those should likely be addressed in a separate PR and is not caused by you in any way. Just unfortunate timing.

Drakulix:
> Also has the CI changed? It seems like clippy is complaining about things that were already there...

Should be fixed in https://github.com/Smithay/drm-rs/pull/122, so after a rebase CI should be good whenever you find the time to work on this again. (No rush!)

TheEdda:
I've rebased on top of current `develop` and used `bytemuck` git dependency again. Unfortunately this doesn't pull in the `bytemuck-derive` git dependency because that crate is not in the root of the repo - and that is the one that needs to update for `check-minimal` to pass. I'll wait until Lokathor is kind enough to release that crate and then finally I hope this PR will be ready.

MarijnS95:
> Unfortunately this doesn't pull in the `bytemuck-derive` git dependency because that crate is not in the root of the repo

And it doesn't use a `[workspace] members = [...]` at the root to include it, so that a `bytemuck-derive = { git = "..." }` would automatically find it? :disappointed:

TheEdda:
@MarijnS95 I didn't know about that feature, but unfortunately no it doesn't. The root Cargo.toml is for the `bytemuck` crate and then there is a `derive` subfolder for the derive crate, but the root Cargo.toml doesn't link to it.

This PR has to wait for a crates.io release anyway so I don't think it's a huge issue. I did test the minimal check locally and it passed with the changes that were merged so I'm hoping nothing breaks.

MarijnS95:
@TheEdward162 Perhaps nice to contribute that to them. The crate is already implicitly included (in i.e. CI linting and testing) because it's a `[dependencies]` of the root crate, but having it in the `[workspace]` makes it even better :partying_face: 

EDIT: Yes, regardless, this (`drm-rs`) crate can't release with a `git` dependency inside of it.

Drakulix:
Can we advance this PR with the newly released bytemuck 1.11? Or are we waiting for more?

TheEdda:
Ah yes, sorry, I missed that release 🤦‍♀️ Should be everything now...

TheEdda:
Will this ever end?

TheEdda:
Since https://github.com/Smithay/drm-rs/pull/124 was merged I've squashed all the commits and cleaned up conflicts. Still waiting for that bytemuck release.

I'm interested in your thoughts about this: https://github.com/Smithay/drm-rs/pull/119/commits/a6237d1983d8826d39965f2106f82456b99445ee#diff-4007187965c1fea4fd252a76f5f1007b2ea59f6d86e2f454edee547c5036be24R10

I only added it to debug builds but maybe it could go into release builds as well, as a sanity check?

TheEdda:
Whee, that should be everything now - please review

Drakulix (approved):
Looks good to me. Thanks a bunch! :)

PR #121, opened 2022-07-15
connector: Don't always do implicit force probe

Drakulix:
The kernel docs state (as easily viewable [here](https://dri.freedesktop.org/docs/drm/gpu/drm-uapi.html#c.drm_mode_get_connector)):

> To retrieve the number of elements, set count_props and count_encoders to zero, set count_modes to 1, and set modes_ptr to a temporary [struct drm_mode_modeinfo](https://dri.freedesktop.org/docs/drm/gpu/drm-uapi.html#c.drm_mode_modeinfo) element.
>
> Performing the ioctl only twice may be racy: the number of elements may have changed with a hotplug event in-between the two ioctls. User-space is expected to retry the last ioctl until the number of elements stabilizes. The kernel won’t fill any array which doesn’t have the expected length.

> If the count_modes field is set to zero and the DRM client is the current DRM master, the kernel will perform a forced probe on the connector to refresh the connector status, modes and EDID. A forced-probe can be slow, might cause flickering and the ioctl will block.
> 
> User-space needs to force-probe connectors to ensure their metadata is up-to-date at startup and after receiving a hot-plug event. User-space may perform a forced-probe when the user explicitly requests it. User-space shouldn’t perform a forced-probe in other situations.

drm-rs currently always performs an implicit force probe (as the first ioctl call uses a mode-length of 0) *and* also does at most two ioctl calls.

This PR aims to fix both of these. `get_connector` now has a `force_probe` flag to toggle calling the function with a mode-length of 0 or 1. Additionally the second ioctl is repeated until the amount of modes is stable. This should avoid races when enumerating a new connector in reaction to a hot-plug event.

### DIFF
--- a/drm-ffi/src/mode.rs
+++ b/drm-ffi/src/mode.rs
@@ -357,13 +357,69 @@ pub fn move_cursor(fd: RawFd, crtc_id: u32, x: i32, y: i32) -> Result<drm_mode_c
 pub fn get_connector(
     fd: RawFd,
     connector_id: u32,
-    props: Option<&mut &mut [u32]>,
-    prop_values: Option<&mut &mut [u64]>,
+    mut props: Option<&mut Vec<u32>>,
+    mut prop_values: Option<&mut Vec<u64>>,
     mut modes: Option<&mut Vec<drm_mode_modeinfo>>,
-    encoders: Option<&mut &mut [u32]>,
+    mut encoders: Option<&mut Vec<u32>>,
+    force_probe: bool,
 ) -> Result<drm_mode_get_connector, Error> {
-    let modes_count = if modes.is_some() {
+    assert_eq!(props.is_some(), prop_values.is_some());
+    let (mut count_modes, mut count_encoders, mut count_props) =
+        if modes.is_some() || encoders.is_some() || props.is_some() {
+            let tmp_mode_info = force_probe.then(|| [drm_mode_modeinfo::default()]);
+            let mut info = drm_mode_get_connector {
+                connector_id,
+                modes_ptr: map_ptr!(&tmp_mode_info),
+                count_modes: if force_probe { 0 } else { 1 },
+                ..Default::default()
+            };
+
+            unsafe {
+                ioctl::mode::get_connector(fd, &mut info)?;
+            }
+
+            (info.count_modes, info.count_encoders, info.count_props)
+        } else {
+            (0, 0, 0)
+        };
+
+    let info = loop {
         let mut info = drm_mode_get_connector {
+            encoders_ptr: match encoders.as_mut() {
+                Some(encoders) => {
+                    encoders.clear();
+                    encoders.reserve_exact(count_encoders as usize);
+                    encoders.as_ptr() as _
+                }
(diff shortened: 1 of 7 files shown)


drm-ffi/src/mode.rs
```diff
@@ -377,28 +381,36 @@ pub fn get_connector(
...
+        };
 
-    unsafe {
-        ioctl::mode::get_connector(fd, &mut info)?;
-    }
+        unsafe {
+            ioctl::mode::get_connector(fd, &mut info)?;
+        }
+
+        if info.count_modes == modes_count {
```
ids1024: The test for this in [libdrm](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/xf86drmMode.c) is:

```c
	if (counts.count_props < conn.count_props ||
	    counts.count_modes < conn.count_modes ||
	    counts.count_encoders < conn.count_encoders) {
```

Drakulix:
@Slabity I would like a review on this, if you got a quick minute.

I have changed the code given the problems with the current approach to rely more an `Vec`s for dynamic allocation. Given that the function used dynamic allocations before, I don't think this change is rather controversial.

But I am personally not a big fan of relying on assumptions about the amount of values the kernel is going to send us anywhere, given that is not defined and thus not part of the public ioctl api. Which is why I would propose to give functions working similarly to `get_connector` inside drm-ffi the same treatment and thus rely more on dynamic allocations overall to avoid running into similar errors again on newer kernel versions. (The amount of properties is only increasing as well, not just encoders and modes.)

In #119 you said you are less concerned about allocation given that custom-allocators exist now, but I am not sure to which extend, given that this was supposed to be a core-compatible crate. With more use of vectors we would effectively require `alloc` for drm-ffi, but I don't see how we could make a sane api otherwise, except for more closely mimicing the ioctl api and require the users to query the amount before and allocating the slice themselves. But that wouldn't feel very ergonomic either.

Any opinions on the general direction you would want to go on this issue?

Slabity:
When I first wrote the code I didn't want to do it because there was no custom allocator support and Rust still used `jemalloc`, which could cause some very annoying to debug issues when integrated with other programs/libraries that used the system allocator.

Now that custom allocator support is integrated and Rust uses the system allocator as default, I am 100% fine (and would actually prefer) that we use `Vec` to replace those. I think you already understand the reasons why :laughing: 

Two things I would like us to do though:

1. Use `alloc::vec::Vec` instead of the one in `std`. I'd like to be able to make this crate `no_std` eventually, and now that the `nix` crate supports it, I think that we can do that relatively easily. I'll take a look at doing that soon.
2. Initialize the `Vec` objects with decently sized-lengths when created. Either using the current sizes we have for the arrays, or even larger. Each time we fail to allocate a large enough buffer, we need to make a second `ioctl` system-call, which I'd like to reduce as much as possible, especially if it's part of a program's fast path. The less time spent in kernel-space the better, and I'm not really concerned about the over-allocation of a couple of bytes.

I haven't looked at the PR yet, but I'll check it out shortly.

Drakulix:
Superseded by #124

PR #124, opened 2022-07-27
Use dynamic allocations

Drakulix:
After the discussion in #121 we concluded, that the current approach of using fixed length static slices for querying dynamic resources inside drm-rs is suboptimal and error-prone.

Commit 1 replaces all questionable uses of this pattern with `Vec`s. This results in some code being simplified.

Commit 2 is just exposing to variables we always queried, but were never readable by the user.

Commit 3 supersedes #121 by fixing the original issue described in there.

@Slabity A review on the first commit would be nice, given the amount of changes. I am feeling fairly certain with the rest.

### DIFF
--- a/drm-ffi/src/lib.rs
+++ b/drm-ffi/src/lib.rs
@@ -76,12 +76,20 @@ pub mod auth {
 }
 
 /// Load this device's Bus ID into a buffer.
-///
-/// If the buffer is too small, this will load the maximum bytes in the buffer.
-/// If the buffer is too big, this will coerce the buffer to the proper size.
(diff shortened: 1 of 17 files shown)


Drakulix:
@Slabity gentle ping. :)

Slabity:
Hey, sorry. I will definitely take time to look at this tonight. A lot of distractions over the past few weeks.

Slabity (approved):
Everything looks good to me. I haven't tested it, but I'm assuming you did on your end?

Drakulix:
Yes, I did test the examples extensively.

Also I won't immediately release this on crates.io after merge, but do a draft update for smithay and cosmic-comp and drive the compositor with the new changes for a couple of days. After that I will be very confident, that I didn't break anything accidentally.

PR #128, opened 2022-09-02
drm-ffi/sys: Regenerate FreeBSD bindings

ids1024:
Generated using the `update_bindings` feature on FreeBSD 13.1. The crate seems to work with this. In an case, it was entirely broken previously since the generated bindings used `::std::` while the crate is `no_std`.


MarijnS95:
Yeah there was previously a PR for this, but the author _and that PR #88_ seem to have disappeared entirely, this is the only reference still pointing to it: https://github.com/Smithay/drm-rs/pull/114#issuecomment-1083637590

PR #130, opened 2022-09-13
Also use add_fb2 for packed buffers. Allows to pass pixel_format.

Matthias-Fauconneau:
Hello,
I'm experimenting with getting 10bit HDR passthrough on my laptop.
It looks like I needed to use add_fb2 to get the 10bit pixel format through.
This patch makes sense to me, but I'm a newbie to this so your mileage may vary.
Cheers

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -266,21 +266,23 @@ pub trait Device: super::Device {
     fn add_framebuffer<B>(
         &self,
         buffer: &B,
-        depth: u32,
-        bpp: u32,
+        modifier: Option<DrmModifier>,
+        flags: u32,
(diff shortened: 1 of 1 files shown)


Drakulix:
There is already a function called `add_planar_framebuffer`, that internally uses `add_fb2`. Why can't you implement `PlanarBuffer` for your buffer instead of changing `add_framebuffer`? Generally we cannot expect `add_fb2` to be available, that depends on the driver in questions. (See https://docs.rs/drm/latest/drm/enum.DriverCapability.html#variant.AddFB2Modifiers.)

Matthias-Fauconneau:
Looks like it could work. Maybe add a note to the documentation then, since it's not obvious that's the way to pass pixel format.

PR #132, opened 2022-11-14
Use `AsFd` instead of `AsRawFd`

ids1024:
This raises the minimum Rust version to 1.63.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,11 @@ trait. Here is a basic example using `File` as a backend:
 /// A simple wrapper for a device node.
 pub struct Card(std::fs::File);
 
-/// Implementing `AsRawFd` is a prerequisite to implementing the traits found
-/// in this crate. Here, we are just calling `as_raw_fd()` on the inner File.
+/// Implementing `AsFd` is a prerequisite to implementing the traits found
+/// in this crate. Here, we are just calling `as_fd()` on the inner File.
 impl std::os::unix::io::AsRawFd for Card {
-    fn as_raw_fd(&self) -> std::os::unix::io::RawFd {
-        self.0.as_raw_fd()
+    fn as_fd(&self) -> std::os::unix::io::BorrowedFd<'_> {
+        self.0.as_fd()
     }
 }
 
--- a/drm-ffi/drm-sys/build.rs
+++ b/drm-ffi/drm-sys/build.rs
@@ -143,7 +143,7 @@ mod use_bindgen {
         println!("cargo:rerun-if-changed={}", dest_file.display());
 
         fs::create_dir_all(&dest_dir).unwrap();
-        fs::copy(&bind_file, &dest_file).unwrap();
+        fs::copy(bind_file, &dest_file).unwrap();
 
         if let Ok(github_env) = var("GITHUB_ENV") {
             let mut env_file = fs::OpenOptions::new()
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -31,11 +31,11 @@ trait. Here is a basic example using `File` as a backend:
 /// A simple wrapper for a device node.
 pub struct Card(std::fs::File);
 
-/// Implementing `AsRawFd` is a prerequisite to implementing the traits found
-/// in this crate. Here, we are just calling `as_raw_fd()` on the inner File.
+/// Implementing `AsFd` is a prerequisite to implementing the traits found
+/// in this crate. Here, we are just calling `as_fd()` on the inner File.
```
MarijnS95, line 35: Making these and the documentation below intradoc links might be nice?

Slabity:
So just to make sure I'm understanding this correctly, the difference between `AsRawFd` and `AsFd` is that the former does not guarantee the file descriptor is valid (like the difference between a `*const T` and `&T`). The literal difference is replacing `RawFd` with `BorrowedFd`. Is this correct?

If that's all, then I have no concerns. Though you will want to ask someone like @Drakulix if the minimum version bump is fine with regards to the other parts of the Smithay project.

ids1024:
> So just to make sure I'm understanding this correctly, the difference between AsRawFd and AsFd is that the former does not guarantee the file descriptor is valid (like the difference between a *const T and &T). The literal difference is replacing RawFd with BorrowedFd. Is this correct?

Yep. It's also unsafe under the IO safety rules to be operating on an fd that may be invalid or (more importantly) may belong to something else. And `OwnedFd`/`BorrowedFd` can be generally handy to make sure an fd is only used while valid and isn't leaked, though in this particular case the ergonomics of implementing `AsRawFd` and `AsFd` are basically the same.

I know the minimum version is a concern in wayland-client/smithay-client-toolkit (which are used in winit), but I don't think that's a concern in smithay itself and components like this used on the compositor side.

Slabity:
> Yep. It's also unsafe under the IO safety rules to be operating on an fd that may be invalid or (more importantly) may belong to something else. And `OwnedFd`/`BorrowedFd` can be generally handy to make sure an fd is only used while valid and isn't leaked, though in this particular case the ergonomics of implementing `AsRawFd` and `AsFd` are basically the same.

That makes sense. An operation on an invalid file descriptor won't do anything unsafe (will just return `EBADFD`), but if a new file descriptor is opened that so happens to have the same number as the previously closed one, then it does have potential to cause an issue.

It will require some small downstream changes in anything that uses this, but that shouldn't be too big of a deal. I'll let the Smithay developers decide whether the version bump is fine. Other projects seemed to be pinned to older versions anyways, so it shouldn't affect them.

Drakulix:
> I'll let the Smithay developers decide whether the version bump is fine.

It's fine, we just bumped smithay ourselves: https://github.com/Smithay/smithay/pull/814

But clippy seems to be complaining @ids1024.

README.md
ids1024: This is a comment in a code example in the `README`, so it can't really be a working link to anything. Similarly with the comment in `examples/` (is there any option that will make `cargo doc` document examples?).

ids1024:
The clippy lints are unrelated to this change and seem to only occur on nightly, but I've added a commit for those. Should pass CI though it shows "First-time contributors need a maintainer to approve running workflows."

README.md
MarijnS95: Isn't the readme synced with top-level library module docs?

MarijnS95: > is there any option that will make `cargo doc` document examples?

Pretty sure `rust-analyzer` can show these, and maybe with `--all-targets` + `--document-private-items`?

PR #129, opened 2022-09-11
Update nix package

electron271:
Related to https://github.com/Smithay/smithay/pull/745

PR #133, opened 2022-11-15
deps: Update nix to 0.25

Drakulix:
Given #132 needs to be released as a new version anyway, lets bump our other dependencies major versions as well.
This will also get rid of the open and stale #129 PR.


Drakulix:
Superseded by #134

PR #134, opened 2022-12-22
Bump `nix` to 0.26

ids1024:


Drakulix:
clippy is failing, but that is not the cause of this PR. We might need to pin our clippy-CI-job to the MSRV.

PR #137, opened 2023-02-21
drm-ffi/sys: Add support for Android platform

rsglobal:
Android use mainline linux kernel DRM/KMS IOCTL API, therefore symlinking to linux platform files should work just fine.


Drakulix (approved):
Well, I guess if for a particular android platform the kernel is severely outdated, downstream can always generate their own bindings with the bindgen feature, so... looks fine by me!

PR #138, opened 2023-02-24
Document `force_probe` parameter on `get_connector()`

PolyMeilex:


src/control/mod.rs
```diff
@@ -127,6 +127,16 @@ pub trait Device: super::Device {
     }
 
     /// Returns information about a specific connector
+    ///
+    /// ## Force-probing
+    ///
+    /// If the `force_probe` is set to `true` and the DRM client is the current DRM master,
```
```suggestion
    /// If `force_probe` is set to `true` and the DRM client is the current DRM master,
```

MarijnS95 (approved):
Nit: I'd change the title to: "Document `force_probe` parameter on `get_connector()`", as just "update" isn't very descriptive IMO.

Drakulix (approved):

PR #140, opened 2023-02-25
Add helper methods for property::Value

PolyMeilex:
Helper methods that transform Value enum into Option of given variant.
Easier to chain with other Options/Results, and easier to early returns with `?`.
Inspired by convention often used by serde crates.

### DIFF
--- a/src/control/property.rs
+++ b/src/control/property.rs
@@ -205,6 +205,83 @@ impl<'a> From<Value<'a>> for RawValue {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Drakulix:
Looks like good changes to me. Could you rebase? CI should be fixed now.

PolyMeilex:
> Looks like good changes to me. Could you rebase? CI should be fixed now.

Done

Slabity:
Looks good to me as well.

PR #141, opened 2023-02-25
Add PropertyValueSet iterator

PolyMeilex:


src/control/mod.rs
```diff
@@ -1172,6 +1172,12 @@ impl PropertyValueSet {
     pub fn as_props_and_values(&self) -> (&[property::Handle], &[property::RawValue]) {
         (&self.prop_ids, &self.prop_vals)
     }
+
+    /// Returns iterator over pairs representing a set of [`property::Handle`] and their raw values
+    pub fn iter(&self) -> impl Iterator<Item = (&property::Handle, &property::RawValue)> {
```
MarijnS95, line 1178: Would it make sense to also `impl IntoIterator for &'a PropertyValueSet` returning this `self.iter()`, afaik allowing it to be used in for loops?

PolyMeilex: Added `IntoIterator` for `&PropertyValueSet` and for `PropertyValueSet`

MarijnS95 (approved):

Drakulix (approved):

PR #142, opened 2023-02-25
Introduce Interface::as_str

PolyMeilex:
This is basically drm-rs equivalent of libdrm's [drmModeGetConnectorTypeName](https://gitlab.freedesktop.org/mesa/drm/-/merge_requests/222) which is meant to unify naming of interfaces across compositors.

### DIFF
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -133,6 +133,36 @@ pub enum Interface {
     USB,
 }
 
+impl Interface {
+    /// Get interface name as string
+    pub fn as_str(&self) -> &'static str {
+        // source: https://github.com/torvalds/linux/blob/489fa31ea873282b41046d412ec741f93946fc2d/drivers/gpu/drm/drm_connector.c#L89
+        match self {
+            Interface::Unknown => "Unknown",
+            Interface::VGA => "VGA",
+            Interface::DVII => "DVI-I",
+            Interface::DVID => "DVI-D",
+            Interface::DVIA => "DVI-A",
+            Interface::Composite => "Composite",
+            Interface::SVideo => "SVIDEO",
+            Interface::LVDS => "LVDS",
+            Interface::Component => "Component",
+            Interface::NinePinDIN => "DIN",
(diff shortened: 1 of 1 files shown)


Slabity:
Looks good to me.

Slightly related, but each time we add a new interface, it is *technically* a breaking change, as downstream users that match on the enum (specifically those that do exhaustively match) would run into a `non-exhaustive patterns` compilation error.

I think it might be worth adding `#[non_exhaustive]` as an attribute to indicate more entries will likely be added in the future. This would force users to match properly and prevent further breaking changes in the future. Though that can be done in a separate PR

MarijnS95:
Separate PR for sure. It would've been nice if the two commits in this PR made it in separately too, @Slabity's comment notifications didn't make much sense with just the current title of the PR :)

src/control/connector.rs
```diff
@@ -127,6 +127,39 @@ pub enum Interface {
...
     DSI,
     DPI,
+    Writeback,
+    SPI,
+    USB,
+}
+
+impl Interface {
+    /// Get interface name as string
+    pub fn as_str(&self) -> &str {
```
MarijnS95: This should be `&'static` as you're returning string literals, otherwise the lifetime of the returned `&str` is unnecessarily bound to `&self`.

PolyMeilex: Done

MarijnS95 (approved):
Looks okay now but I think the addition of new fields and migration to non_exhaustive should still happen in a separate PR.

PolyMeilex:
> Looks okay now but I think the addition of new fields and migration to non_exhaustive should still happen in a separate PR.

Here you go: https://github.com/Smithay/drm-rs/pull/143

PolyMeilex:
Converted to draft until https://github.com/Smithay/drm-rs/pull/143 is merged

Slabity:
@PolyMeilex - I have merged #143 . If you can rebase your branch, then this should be good to go.

Slabity:
Looks good to me. Merging now

PR #143, opened 2023-02-27
Update connector Interface enum & make it non_exhaustive

PolyMeilex:


Slabity:
I was a little confused how `From<Interface> for u32` was compiling without the `_` match case, but apparently `#[non_exhaustive]` only has an effect when imported. It doesn't have any effect within its own crate.

I'll go ahead and merge this now. We can merge #142 afterwards

PolyMeilex:
> I was a little confused how `From<Interface> for u32` was compiling without the `_` match case, but apparently `#[non_exhaustive]` only has an effect when imported. It doesn't have any effect within its own crate.
> 
> I'll go ahead and merge this now. We can merge #142 afterwards

yep, non_exhaustive works only in public APIs, anything done internally in the crate is nicely checked. Best of both worlds.

PR #147, opened 2023-04-05
Add connector subpixel getter

PolyMeilex:
Enum definition on kernel side: https://github.com/torvalds/linux/blob/76f598ba7d8e2bfb4855b5298caedd5af0c374a8/include/drm/drm_connector.h#L136

libdrm [offsets this enum by 1](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/xf86drmMode.c#L570) that's why in [xf86DrmMode.h](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/xf86drmMode.h#L196) the values are different.

### DIFF
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -260,6 +260,7 @@ impl From<State> for u32 {
 }
 
 /// Subpixel order of the connected sink
+#[non_exhaustive]
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub enum SubPixel {
     /// Unknown geometry
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -292,19 +292,3 @@ impl From<u32> for SubPixel {
         }
     }
 }
-
-impl From<SubPixel> for u32 {
-    fn from(state: SubPixel) -> Self {
-        // These values are not defined in drm_mode.h
-        // They were copied from linux's drm_connector.h
-        // Don't mistake those for ones used in xf86DrmMode.h (libdrm offsets those by 1)
-        match state {
-            SubPixel::Unknown => 0,
-            SubPixel::HorizontalRgb => 1,
-            SubPixel::HorizontalBgr => 2,
-            SubPixel::VerticalRgb => 3,
-            SubPixel::VerticalBgr => 4,
-            SubPixel::None => 5,
-        }
-    }
-}
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -275,6 +275,8 @@ pub enum SubPixel {
     VerticalBgr,
     /// No geometry
     None,
+    /// Encountered value not supported by drm-rs
+    NotImplemented,
 }
 
 impl SubPixel {
@@ -289,7 +291,7 @@ impl SubPixel {
             3 => Self::VerticalRgb,
             4 => Self::VerticalBgr,
             5 => Self::None,
-            _ => Self::Unknown,
+            _ => Self::NotImplemented,
         }
     }
 }
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -57,6 +57,7 @@ pub struct Info {
     pub(crate) modes: Vec<control::Mode>,
     pub(crate) encoders: Vec<control::encoder::Handle>,
     pub(crate) curr_enc: Option<control::encoder::Handle>,
+    pub(crate) subpixel: SubPixel,
 }
 
 impl Info {
@@ -101,6 +102,11 @@ impl Info {
     /// Returns the current encoder attached to this connector.
     pub fn current_encoder(&self) -> Option<control::encoder::Handle> {
         self.curr_enc
+    }
+
+    /// Subpixel order of the connected sink
+    pub fn subpixel(&self) -> SubPixel {
+        self.subpixel
     }
 }
 
@@ -249,6 +255,56 @@ impl From<State> for u32 {
             State::Connected => 1,
             State::Disconnected => 2,
             State::Unknown => 3,
+        }
+    }
+}
+
+/// Subpixel order of the connected sink
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum SubPixel {
+    /// Unknown geometry
+    Unknown,
+    /// Horizontal RGB
+    HorizontalRgb,
+    /// Horizontal BGR
+    HorizontalBgr,
+    /// Vertical RGB
+    VerticalRgb,
+    /// Vertical BGR
+    VerticalBgr,
+    /// No geometry
+    None,
+}
+
+impl From<u32> for SubPixel {
+    fn from(n: u32) -> Self {
+        // These values are not defined in drm_mode.h
+        // They were copied from linux's drm_connector.h
+        // Don't mistake those for ones used in xf86DrmMode.h (libdrm offsets those by 1)
+        match n {
+            0 => Self::Unknown,
+            1 => Self::HorizontalRgb,
+            2 => Self::HorizontalBgr,
+            3 => Self::VerticalRgb,
+            4 => Self::VerticalBgr,
+            5 => Self::None,
+            _ => Self::Unknown,
+        }
+    }
+}
+
+impl From<SubPixel> for u32 {
+    fn from(state: SubPixel) -> Self {
+        // These values are not defined in drm_mode.h
+        // They were copied from linux's drm_connector.h
+        // Don't mistake those for ones used in xf86DrmMode.h (libdrm offsets those by 1)
+        match state {
+            SubPixel::Unknown => 0,
+            SubPixel::HorizontalRgb => 1,
+            SubPixel::HorizontalBgr => 2,
+            SubPixel::VerticalRgb => 3,
+            SubPixel::VerticalBgr => 4,
+            SubPixel::None => 5,
         }
     }
 }
--- a/src/control/connector.rs
+++ b/src/control/connector.rs
@@ -277,8 +277,8 @@ pub enum SubPixel {
     None,
 }
 
-impl From<u32> for SubPixel {
-    fn from(n: u32) -> Self {
+impl SubPixel {
+    pub(super) fn from_raw(n: u32) -> Self {
         // These values are not defined in drm_mode.h
         // They were copied from linux's drm_connector.h
         // Don't mistake those for ones used in xf86DrmMode.h (libdrm offsets those by 1)
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -170,7 +170,7 @@ pub trait Device: super::Device {
             modes: Mode::wrap_vec(modes),
             encoders: unsafe { transmute_vec_from_u32(encoders) },
             curr_enc: unsafe { mem::transmute(ffi_info.encoder_id) },
-            subpixel: connector::SubPixel::from(ffi_info.subpixel),
+            subpixel: connector::SubPixel::from_raw(ffi_info.subpixel),
         };
 
         Ok(connector)
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -170,6 +170,7 @@ pub trait Device: super::Device {
             modes: Mode::wrap_vec(modes),
             encoders: unsafe { transmute_vec_from_u32(encoders) },
             curr_enc: unsafe { mem::transmute(ffi_info.encoder_id) },
+            subpixel: connector::SubPixel::from(ffi_info.subpixel),
         };
 
         Ok(connector)


src/control/connector.rs
```diff
@@ -252,3 +258,53 @@ impl From<State> for u32 {
...
+        match state {
+            SubPixel::Unknown => 0,
+            SubPixel::HorizontalRgb => 1,
+            SubPixel::HorizontalBgr => 2,
+            SubPixel::VerticalRgb => 3,
+            SubPixel::VerticalBgr => 4,
+            SubPixel::None => 5,
+        }
+    }
+}
```
Drakulix: Given we are not using the same definition as `xf86DrmMode`, I am wondering if we should expose this conversion trait at all. The returned values don't have any particular meaning to downstream and there is no place to publically document why these are seemingly shifted by 1.

Drakulix (commented):
Looks good! Just one nitpick

src/control/connector.rs
```diff
@@ -252,3 +258,53 @@ impl From<State> for u32 {
...
+        // They were copied from linux's drm_connector.h
+        // Don't mistake those for ones used in xf86DrmMode.h (libdrm offsets those by 1)
+        match n {
+            0 => Self::Unknown,
+            1 => Self::HorizontalRgb,
+            2 => Self::HorizontalBgr,
+            3 => Self::VerticalRgb,
+            4 => Self::VerticalBgr,
+            5 => Self::None,
+            _ => Self::Unknown,
```
MarijnS95: Feels like this should be an `Err` via a `try_from()`, so that if any new geometries were ever added they aren't treated the same as the known value `0`?

Slabity: If that's the case, we should probably do the same as #143 and make it `#[non_exhaustive]` to allow the ability for new formats (and since it's a new type, no breaking change is made :+1: )

Also, if we're getting these values directly from the kernel, then it might be simpler to make the enum `#[repr(u32)]` and set the values directly in the definition. We'd still need to check the value when we convert to the type, but would provide some guarantees if we try to use it in some other way.

As far as returning `Self::Unknown` vs `Err` (in the case of a new format being added to the kernel), I don't think the `Err` is useful downstream unless we also include the original, unknown value. And if their code checks that value and tries to act on it in some way, then adding the new format would result in being a breaking change for the downstream user, as they would no longer be receiving the `Err` at all to act on. So I'd recommend sticking with `Self::Unknown` unless there's another reason.

src/control/connector.rs
Slabity: If we use `#[repr(u32)]` for this, downstream can simply cast it as `u32` without needing this trait.

On the other hand, maybe these values could vary between different kernels? If that's the case, we would probably want these values to be put into `drm-ffi` and pull them in here.

src/control/connector.rs
MarijnS95: > Also, if we're getting these values directly from the kernel, then it might be simpler to make the enum `#[repr(u32)]` and set the values directly in the definition.

Are there more enums already in this crate that could/should get the same treatment? Perhaps `enum Interface` that was updated in #143?

> unless we also include the original, unknown value.

The `enum SubPixel` is `Copy` so the user can keep it around and reuse it in a hypothetical `Err(UnknownValue)`. I think that's better than treating `0` _and anything `>= 6`_ as the same `Unknown` enum value, as they'll all translate back to `0`.

Slabity: > Are there more enums already in this cratre that could/should get the same treatment? Perhaps enum Interface that was updated in https://github.com/Smithay/drm-rs/pull/143?

Yes, there's probably a bunch of these actually. I was unaware of `#[repr(u*)]` and `#[repr(i*)]` when the crate was initially created, so it might be a good idea to go through all of those at some later point.

> The `enum SubPixel` is `Copy` so the user can keep it around and reuse it in a hypothetical `Err(UnknownValue)`.

The only concern I have about this is that it could be a potential cause of breakage for downstream users if/when a new format gets added.

Let's say a kernel adds a new hypothetical subpixel format `HorizontalRGBY` with a value of `6`. Our code would return `Err(6)` to tell downstream users, "This is a valid format according to the kernel, but we don't have it hardcoded in our crate yet". The downstream user sees this and can do one of two things:

1. Treat it as some unknown format and don't do any subpixel rendering (safe, but no different than receiving a `Subpixel::Unknown`)
2. Look up the format on their side, see `Err(6)` means `HorizontalRGBY`, and then do some sort of subpixel formatting based on that information.

The problem with the second path is what happens when we add `Subpixel::HorizontalRGBY` to the crate and return that instead? We are now no longer returning `Err(6)` and downstream's path for this format is broken until they update their end to use the proper enum variant.

MarijnS95: @Slabity I wasn't actually suggesting to use `Err(6)` directly (the `UnknownValue` could be a ZST struct or enum value without contents), and the user can simply map the original value in that case since `SubPixel` is `Copy`. Thanks to `non_exhaustive` their code should also catch new unhandled `enum` values via however way they wish to handle it.

After all this would be a kernel API change/extension, and the only thing drm-rs can do about this is communicating it as transparently to the user as possible, without possibly giving the ambiguity of having `Err(6)` at some point and a `HorizontalRGBY` later.

src/control/connector.rs
PolyMeilex: Removed all together, I agree with Drakulix, I believe that using values different than xf86DrmMode would only confuse users.

src/control/connector.rs
PolyMeilex: Made the whole thing private, subpixel() getter will still return unknown as is did, if we want to change the way we handle enums I would do that in all enums at the same time.

In addition to that, as stated in the second thread, I believe using values different than xf86DrmMode here would only confuse people, so keeping it private may simply be the best choice.

PolyMeilex:
I made it #[non_exhaustive] as well

src/control/connector.rs
MarijnS95: Still disagree about mapping `0` **and `> 5`** to `Unknown`, making the user unable whatsoever to intercept new values unlike an `UnknownValue(u32)`, but alas.

Drakulix: I agree, that the Unknown variant could have an integer value, but then all of them should sooo...

Given it is not_exhaustive, I feel like this is good to merge until we made a definite decision on this type of enums for the whole project? @Slabity any objections?

I tend to agree with @Slabity here in the general case though. I think we should avoid breakage and are quick to update the crate anyway, if anyone wants to add new variants, so keeping this up-to-date doesn't seem like an issue to me.

MarijnS95: > I think we should avoid breakage and are quick to update the crate anyway

@Drakulix This, in conjunction with `non_exhaustive` makes me prefer a `todo!("Unknown value {}")`. That way:

- The user is informed about the unknown value;
- There's no ambiguity behind `Unknown` being `0` or `> 5`;
- There's no ambiguity between `Err(6)` and a possible new `HorizontalRGBY = 6,` variant;
- An issue/PR (for visibility!) and crate update are more likely (since the crate cannot be used without a fix).

Drakulix: But wouldn't that mean a new kernel (returning a new value) could potentially break existing binaries? That doesn't sound too great to me.

MarijnS95: Correct.

Do we expect to extend this enum value often? If not, perhaps it is safer to drop `non_exhaustive` and have an explicit `UnknownValue(u32)` (for all enums across this crate if it makes things consistent). That removes ambiguity, as when we add the theoretical `HorizontalRGBY = 6` in a breaking release downstream code is required to extend their `match` statement rather than having this end up in the catch-all arm that they are forced to have?

Drakulix: > Do we expect to extend this enum value often?

No we do not, which is why I think updating drm-rs once new values actually get added is reasonable from a maintenance perspective.
Which means we could have a stable api and stable binaries at the cost of having to extend this enum, if it ever gets updated.

Which seems to be the better trait off to me than either having binaries panic on newer kernels or having matches on `Unknown`-variants with values break in the future.

> If not, perhaps it is safer to drop non_exhaustive and have an explicit UnknownValue(u32) (for all enums across this crate if it makes things consistent).

That would also be an option that satisfies these constraints. The only immediate benefit I see in this approach is, that it doesn't force users to upgrade to newer drm-rs releases to handle new values. Which might be nice to handle compatibilities with other crates or rust versions, though we don't frequently bump the MSRV and most accompanying crates (like gbm.rs) have a release almost immediately following drm-rs.

So the solutions are either:
- Opaque enum (raw values inaccessible including) and non-exhaustive.
- Visible unknown values and exhaustive.

The latter is more flexible, but requires more frequent major version bumps by semver. Which is why I slightly prefer the former, but either is fine to me.

Drakulix: @Slabity would you mind making a decision here? I think all pros and cons have been clearly layed out.

MarijnS95: @Drakulix yeah, thanks for laying out the two options more explicitly and concisely.

I think I prefer the first option with `non_exhaustive` and an opaque unknown variant as well, as long as it's different from `Unknown = 0`. That both enables `drm-rs` to do frequent _minor_ bumps, as well as force crate authors to update often if they wish to handle this case by not giving them access to the value.

It has been a while since I used `non_exhaustive` however: is there already a lint that is able to tell you if there are new variants in the `enum` that a user does not match on, because they are forced to have a catch-all `_ => {...}` arm? To give more visibility into such changes in updates.

Drakulix: > It has been a while since I used non_exhaustive however: is there already a lint that is able to tell you if there are new variants in the enum that a user does not match on, because they are forced to have a catch-all _ => {...} arm? To give more visibility into such changes in updates.

If there isn't that is something we should open a `clippy`-issue on.

Drakulix: > I think I prefer the first option with non_exhaustive and a non-opaque unknown variant as well, as long as it's different from Unknown = 0

If the unknown variant is non-opaque we run into the same problem again. Either everything is opaque and Unknown can't be differentiated from a legitimately unknown variant + non_exhaustive **or** nothing is opaque and it's also exhaustive.

MarijnS95: @Drakulix whoops I meant to ~say~ quote "opaque", but indeed we'd still have to come up with a plausible discriminant for it that doesn't overlap with a possible kernel value in the future - and that's exactly impossible when the kernel value is `u32` and the `enum` is `repr(u32)` too.

Unless the raw/discriminant values are removed as you suggested above.

Drakulix: Right, which is why the enum can't be `repr(u32)`. We can add a new variant `Invalid`, if you want to explicitly differentiate between "unknown value" and "unknown as a value".

MarijnS95: In one of the crates that I maintain where we use `num_enum`, we're currently debating a similar issue. There's now a `num_enum(default)` for such an `Invalid` variant, and a `num_enum(catch_all)` for an `Invalid(u32)` variant. This does blow up the `enum` as it generates every variant _with discriminant_ (and uses that to cast everything _but_ the `Invalid` variant), meaning extra space is needed to store the `u32` for `Invariant(u32)`.

I'm still on the fence there which way to go :sweat_smile:

Drakulix: As I said I don't want to expose the raw value, if we go `non_exhaustive`, which I would prefer as an api over an exhaustive variant. I am thus not sure, that an opaque `Invalid` variant has any advantage over just returning `Unknown`, because frankly that is how you should treat the value anyway in that case.

Slabity: Sorry guys, I was away camping for a bit and didn't see the pings.

The major issue I see with using a variant that returns the raw value is that once we *do* add the value properly, it will break everyone's code that uses that variant regardless. If a user matches on something like `Invalid(11)` (or whatever it would be called) and acts on it in some way, then we add `11 => Self::QuantumSuperpositionRGBXYZ` to the enum, then that code path *they* added will break. And this can potentially break *every time we add one*. There's a few different things we can do to get around that, but honestly if someone needs that, then it likely wouldn't be too big of a deal to have them file an issue or open a PR and add a line to fix it.

On the other hand, I don't really want to have people wait for an unknown amount of time before we update the crate to match new values that get implemented upstream. It could get slightly annoying to have these values get implemented from kernel -> `x86drm*` bindings -> `drm-rs` bindings. It won't happen often in this case, but there's other cases where this might potentially happen much more often.

Instead, maybe it would be feasible to try and get these automated? There's not really a lot of human complexity to just add a new value to an enum. It would take some effort at first, but would allow people to update it on their own *and* we'd be able to leverage CI to make maintaining these sort of changes way easier.

Slabity:
I think the PR looks good as it is now. We should probably merge this so that downstream can take advantage of the new pixel layout and not need to wait much longer.

@Drakulix - Looks like a new commit was pushed a few hours ago. If everything looks good to you as well then feel free to merge and we can decide what should be done in the longer term.

src/control/connector.rs
MarijnS95: No worries.

Again, I agree that having an opaque value prevents ambiguity, by _not allowing a user to match on `Invalid(11)`_.

> but honestly if someone needs that, then it likely wouldn't be too big of a deal to have them file an issue or open a PR and add a line to fix it.

Again^2, I also think that mapping **both** `0 => Unknown` **and** `_ => Unknown` actively stalls end-users in reporting such issues. They now have no way (without casting or inspecting structures?) to tell apart whether:

1. The kernel returned `0`;
2. The kernel returned **non-`0`** and I need to ask `drm-rs` to map the new value.

That's why I suggested to cook up a different variant (e.g. `Invalid`) for the `_` case.

Slabity: Ah, sorry, I misinterpreted. I understand now. By adding `Invalid` as its own value (different from the kernel's `Unknown`) we it can be caught downstream and get people to report it more quickly.

That makes a lot of sense, I think that's probably a good idea so long as we document it as such to tell downstream users something like, "Hey, this means *we* don't know what it is yet, not that your system doesn't"

I agree with adding that. Though maybe not calling it `Invalid` as that might make people think the kernel is sending an actual invalid value. Maybe `Unimplemented`?

MarijnS95: Yeah I wasn't too happy about the `Invalid` brought up above; I do like the suggested `Unimplemented` together with an appropriate doc-comment.

Just in case (the outcome of) this long trail of discussion should probably end up in some doc or guideline as it transcends this PR and the simplicity of Subpixel in general. I'm taking these points into account on other crates as well.

Slabity: Sounds good. I won't bikeshed too much on what to call it, but either `Unimplemented` or `NotImplemented` would be fine by me. The latter *might* make it be less associated with the `unimplemented!` macro but honestly I'm not too picky.

Drakulix (approved):

PR #150, opened 2023-07-21
Include structures from xf86drm.h in ffi

russelltg:
I found myself needing to call `drmGetDeviceNameFromFd2` from my app, but realized these functions were missing from drm-sys. 

I'm opening this as a draft to see if this is interesting, and if so I'm happy to generate the bindings for the other platforms as well.


linkmauve:
Hi, this function is a helper from libdrm, which this crate doesn’t use (except for generating the headers).  I would recommend rewriting [this function](https://gitlab.freedesktop.org/mesa/drm/-/blob/6abc164052e4902f67213baa279d743cf46227d4/xf86drm.c#L4822) in Rust rather than including the whole libdrm in this crate.

Slabity:
I think I'm going to close this as it's more related to functionality on `sysfs`.

And once I get traction on #165, I'll likely be removing `libdrm` as a dependency entirely.

PR #151, opened 2023-07-21
link to libdrm in drm-sys

russelltg:
It seems to be typical for `-sys` crates to link to the library they generate bindings for. 

sources: 
https://github.com/alexcrichton/curl-rust/blob/main/curl-sys/build.rs#L530
https://github.com/zmwangx/rust-ffmpeg-sys/blob/master/build.rs#L635

I can also use pkg_config if that's preferred, but currently that's behind a feature flag so I figured I'd keep it that way, plus any sane configuration should have libdrm in default search paths...right....?

### DIFF
--- a/drm-ffi/drm-sys/build.rs
+++ b/drm-ffi/drm-sys/build.rs
@@ -188,4 +188,6 @@ pub fn main() {
             target_os, target_arch
         );
     }
+
+    println!("cargo:rustc-link-lib=drm");
 }


Slabity:
Hey @russelltg, the `drm-sys` crate isn't really an interface to `libdrm`, but rather an interface to the kernel interface directly. The crate only uses `libdrm` to generate bindings because it's *far* easier to do it that way than to try and extract the C structs and `ioctl` commands from the kernel code itself.

Basically, you don't need to actually link to `libdrm` at all. It's only for generating bindings.

russelltg:
I see--so this repo isn't the right place for this or #150 then?

Drakulix:
> I see--so this repo isn't the right place for this or https://github.com/Smithay/drm-rs/pull/150 then?

No not really. But that doesn't mean that we could provide similar functionality to `drmGetDeviceNameFromFd2` and other similar functions from libdrm.

We have use for these in smithay as well and ended up re-implementing a bunch of that logic: https://github.com/Smithay/smithay/blob/master/src/backend/drm/node/mod.rs

@Slabity Do you see that in scope for drm-rs?

Slabity:
> @Slabity Do you see that in scope for drm-rs?
Sorry, I missed this ping.

I'm not opposed to reimplementing functions that are currently in `libdrm`, but looking at how it's implemented in that library, this function only seems to work on Linux. Is it possible to implement it in a way that works on all systems?

Drakulix:
> > @Slabity Do you see that in scope for drm-rs?
> > Sorry, I missed this ping.
> 
> I'm not opposed to reimplementing functions that are currently in `libdrm`, but looking at how it's implemented in that library, this function only seems to work on Linux. Is it possible to implement it in a way that works on all systems?

I don't think the enumeration can be system-agnostic, if we re-implement it and not use some common api like libdrm.

Drakulix:
Given the original intend of this PR isn't going to happen, I am going to close this. But feel free to open an issue regarding reimplementing `drmGetDeviceNameFromFd2` or lets continue discussion in #150.

PR #153, opened 2023-07-28
Add public function to build Event iterator

CirrusNeptune:
If a read function other than `Device::receive_events()` is desired (e.g. tokio AsyncFd), the Event iterator is still usable this way.


Drakulix (approved):
Seems reasonable.

Drakulix:
Thanks!

PR #156, opened 2023-08-12
Use core::ffi instead of libc for C types

notgull:
The goal is to eventually totally get rid of libc and move to direct syscalls via rustix. For now this at least eliminates a dependency from drm-sys.

This requires an MSRV bump to 1.64

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
           sudo apt-get install -y libdrm-dev
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.63.0
+          toolchain: 1.64.0
           profile: minimal
           components: clippy
           default: true
@@ -80,12 +80,12 @@ jobs:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ runner.os }}-cargo-rust_1.63.0-${{ hashFiles('**/Cargo.toml') }}
+          key: ${{ runner.os }}-cargo-rust_1.64.0-${{ hashFiles('**/Cargo.toml') }}
       - name: Build cache
         uses: actions/cache@v2
         with:
           path: target
-          key: ${{ runner.os }}-build-rust_1.63.0-check-${{ hashFiles('**/Cargo.toml') }}
+          key: ${{ runner.os }}-build-rust_1.64.0-check-${{ hashFiles('**/Cargo.toml') }}
       - name: Clippy
         uses: actions-rs/clippy-check@v1
         with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ version = "0.9.0"
 license = "MIT"
 authors = ["Tyler Slabinski <tslabinski@slabity.net>", "Victoria Brekenfeld <crates-io@drakulix.de>"]
 exclude = [".gitignore", ".github"]
-rust-version = "1.63"
+rust-version = "1.64"
 resolver = "2" # Required to separate dev-dependencies.nix features
 
 [dependencies]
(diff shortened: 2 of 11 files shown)


Slabity:
Not going to lie, this is the first time I have heard of the `rustix` crate, though that is likely an issue with me not staying up to date with the wider rust ecosystem.

I am all for reducing the number of run-time link dependencies though, so unless anyone else has an issue with replacing `libc` with `rustix` then I'm not too concerned.

Would this also eventually replace the `nix` crate eventually? Specifically for calling `ioctl`, `mmap`, and a couple of other calls?

ids1024:
> Would this also eventually replace the nix crate eventually?

`rustix` is basically an alternative for `nix`. But makes use of `OwnedFd`/`BorrowedFd`/etc. in APIs. And has optional support for making direct system calls on Linux instead of going through libc.

The way `RawFd` is used in safe functions in `nix` violates Rust's new IO safety rules, so it's probably best to avoid that. Unless `nix` updates to correctly conform to IO safety, but that doesn't seem to be happening too imminently anyway.

So I'd generally agree with moving to `rustix`. If this uses any `nix` APIs that aren't in `rustix`, it should be possible to add those to `rustix`.

notgull:
How do I update the bindings for all platforms, by the way? I only updated them for x86_64 Linux.

ids1024:
I think they're all generated by running on that platform. Which isn't really ideal.

I'm not sure the duplication for platform bindings is really necessary. Looks like `linux/aarch64/bindings.rs` is the same as `linux/x86_64/bindings.rs` and `linux/arm/bindings.rs` is the same as `linux/x86/bindings.rs`. And the small difference is probably from the version of libdrm/libc it was generated from. FreeBSD has a few more differences, but it's all in type aliases that I'm not sure are actually used anywhere.

Slabity:
> I think they're all generated by running on that platform. Which isn't really ideal.
> 
> I'm not sure the duplication for platform bindings is really necessary. Looks like `linux/aarch64/bindings.rs` is the same as `linux/x86_64/bindings.rs` and `linux/arm/bindings.rs` is the same as `linux/x86/bindings.rs`. And the small difference is probably from the version of libdrm/libc it was generated from. FreeBSD has a few more differences, but it's all in type aliases that I'm not sure are actually used anywhere.

Yea, it's not ideal but since we cannot guarantee that the `ioctl` numbers are identical between different kernels or even different architectures, linking them together can cause some difficult to debug errors down the line..

What I'd like to do is set up a Github process to automatically pull and update the bindings. If we can somehow do a "cross-compile" with `bindgen` to automatically use the correct pre-processor macros for each system, then we could have all bindings get updated at once without needing to run the builder on different systems. Though that's likely going to be a more involved process.

Slabity:
I'm going to let someone from the Smithay project approve this one since it updates the MSRV and that doesn't affect me at all.

That said, if this does get approved I will likely look into replacing all the `nix` crate calls with `rustix` at some point and remove that dependency entirely.

notgull:
> That said, if this does get approved I will likely look into replacing all the `nix` crate calls with `rustix` at some point and remove that dependency entirely.

I've worked on https://github.com/bytecodealliance/rustix/pull/788, which should make this possible.

Drakulix:
MSRV bump to 1.64 is fine, smithay currently [tests against 1.65](https://github.com/Smithay/smithay/blob/master/.github/workflows/ci.yml#L89).

notgull:
@Slabity What is the status of this PR?

Slabity:
> @Slabity What is the status of this PR?

Ah, sorry. I was going to let someone from the Smithay project merge this. Since @Drakulix doesn't have any concerns I'll merge this now.

PR #158, opened 2023-09-02
Bump all of the outdated dependencies

linkmauve:
Each commit describes the changes done for this particular bump.

The most interesting one is nix having started to convert to the `OwnedFd`/`BorrowedFd` API, which properly encodes the lifetime of the file descriptors.  This PR doesn’t update drm to make use of these types, but we should eventually do that.


drm-ffi/drm-sys/src/platforms/linux/x86_64/bindings.rs
```diff
@@ -1145,12 +1154,6 @@ pub struct drm_color_lut {
 }
 #[repr(C)]
 #[derive(Debug, Default, Copy, Clone, Hash, PartialEq, Eq)]
-pub struct hdr_metadata_infoframe__bindgen_ty_1 {
-    pub x: __u16,
-    pub y: __u16,
-}
-#[repr(C)]
-#[derive(Debug, Default, Copy, Clone, Hash, PartialEq, Eq)]
 pub struct drm_mode_crtc_page_flip {
```
Drakulix, line 1211: Not sure if this is an issue with our CI platform or bindgen, but why does this drop structs?

linkmauve: This seems like a bug in the old bindgen, that struct is part of `struct hdr_metadata_infoframe` but without the rest of that struct it is completely useless.  Why the main struct isn’t part of the generated file, I don’t know, but nothing can regress from this upgrade.

Drakulix: Ah, this probably happens, because we are [allow/block-listing types to generate bindings for](https://github.com/Smithay/drm-rs/pull/158/files#diff-87619c68d500c3b4ab86e2a26da6c6deb16df1f2d5222b7ebe03757919c3286cR102-R117) and that list doesn't contain the `hdr`-prefix. Would you mind adding that or opening an issue to track that?

Drakulix:
It looks like the nix-update needs rust 1.65. Which is fine, but causes CI to fail. Can you update the toolchain used in the CI script to 1.65?

src/control/mod.rs
```diff
@@ -950,6 +950,7 @@ bitflags::bitflags! {
     /// Limited to the values in [`ffi::drm_sys::DRM_MODE_PAGE_FLIP_FLAGS`],
     /// minus [`ffi::drm_sys::DRM_MODE_PAGE_FLIP_TARGET`] bits which are
     /// passed through [`PageFlipTarget`].
+    #[derive(Clone, Copy, Debug)]
```
MarijnS95: Should these also derive `Eq, PartialEq, Hash`? Besides that only the `Ord` traits are now missing that were present on `bitflags 1` (https://github.com/bitflags/bitflags/releases/tag/2.0.0), but they're not relevant for these types.

Drakulix: Certainly wouldn't hurt.

drm-ffi/drm-sys/src/platforms/linux/x86_64/bindings.rs
linkmauve: Done.

src/control/mod.rs
linkmauve: Done.

linkmauve:
I have added all the changes you requested, they all made sense, thanks!

Drakulix (approved):
Looks great! Thanks a lot :)

PR #160, opened 2023-09-02
Publicly expose the ClipRect type for drm-rs

### LINKED ISSUE #155: Expose drm_sys::drm_clip_rect in the main crate
This is used in the `dirty_framebuffer` function but isn't exposed in the main `drm` crate, necessitating the import of `drm-sys`.

notgull:
Otherwise it is impossible to call the dirty_framebuffer type without also importing drm-sys to get the drm_clip_rect type. Closes #155.

This is a breaking change for `drm-rs`.

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -341,7 +341,10 @@ pub trait Device: super::Device {
         handle: framebuffer::Handle,
         clips: &[ClipRect],
     ) -> Result<(), SystemError> {
-        ffi::mode::dirty_fb(self.as_fd().as_raw_fd(), handle.into(), clips)?;
+        ffi::mode::dirty_fb(self.as_fd().as_raw_fd(), handle.into(), unsafe {
+            // SAFETY: ClipRect is repr(transparent) for drm_clip_rect
+            core::slice::from_raw_parts(clips.as_ptr() as *const ffi::drm_clip_rect, clips.len())
+        })?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


Drakulix:
> This is a breaking change for drm-rs.

Thanks for the heads up, but I think we have accumulated enough unreleased changes, that we can easily do a 0.10 anyway.

src/control/mod.rs
```diff
@@ -341,7 +341,10 @@ pub trait Device: super::Device {
         handle: framebuffer::Handle,
         clips: &[ClipRect],
     ) -> Result<(), SystemError> {
-        ffi::mode::dirty_fb(self.as_fd().as_raw_fd(), handle.into(), clips)?;
+        ffi::mode::dirty_fb(self.as_fd().as_raw_fd(), handle.into(), unsafe {
+            // SAFETY: ClipRect is repr(transparent) for drm_clip_rect
+            core::slice::from_raw_parts(clips.as_ptr() as *const ffi::drm_clip_rect, clips.len())
```
Drakulix, line 346: Since we do now have bytemuck in our dependencies to handle these unsafe operations, would you mind replacing this with a `cast_slice`-call? https://docs.rs/bytemuck/latest/bytemuck/fn.cast_slice.html

To elaborate, I don't think your code is unsafe, but we already shifted that responsibility to bytemuck in other cases, so if we can avoid added unsafe code ourselves here by further relying on that crate, I would like to do that.

Drakulix (commented):
LGTM aside from a small nitpick. Thanks!

src/control/mod.rs
notgull: We can't use `bytemuck` here as `drm_clip_rect` is not marked as `Pod`.

Drakulix: And `drm_clip_rect` is bindgen generated... I see. So lets keep it this way.

Drakulix (approved):

PR #161, opened 2023-09-07
Make add_planar_framebuffer() take only one modifier

linkmauve:
The kernel enforces all four modifiers to actually be the same, so requesting four of them is a legacy artifact by now.  Let’s avoid encoding such legacy into our APIs even more.

Additionally, making `None` default to `LINEAR` doesn’t seem very valid, if there was a default to pick I think it would be `INVALID`, but it is actually much better to ask the user which modifier they actually want.

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -262,11 +262,13 @@ pub trait Device: super::Device {
             Err(UnrecognizedFourcc(_)) => return Err(SystemError::UnknownFourcc),
         };
 
+        let flags = FbCmd2Flags::from_bits_truncate(info.flags);
+
         let fb = framebuffer::PlanarInfo {
             handle,
             size: (info.width, info.height),
             pixel_format,
-            flags: info.flags,
+            flags,
             buffers: bytemuck::cast(info.handles),
             pitches: info.pitches,
             offsets: info.offsets,
@@ -305,7 +307,7 @@ pub trait Device: super::Device {
         &self,
         planar_buffer: &B,
         modifiers: &[Option<DrmModifier>; 4],
-        flags: u32,
+        flags: FbCmd2Flags,
     ) -> Result<framebuffer::Handle, SystemError>
     where
         B: buffer::PlanarBuffer + ?Sized,
@@ -330,7 +332,7 @@ pub trait Device: super::Device {
             &planar_buffer.pitches(),
             &planar_buffer.offsets(),
             &mods,
-            flags,
+            flags.bits(),
         )?;
 
         Ok(from_u32(info.fb_id).unwrap())
@@ -1448,5 +1450,16 @@ bitflags::bitflags! {
         /// older userspace (DDX drivers) that read/write each prop they find,
         /// witout being aware that this could be triggering a lengthy modeset.
         const ATOMIC = ffi::DRM_MODE_PROP_ATOMIC;
+    }
+}
+
+bitflags::bitflags! {
+    /// Planar framebuffer flags
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+    pub struct FbCmd2Flags : u32 {
+        /// For interlaced framebuffers
+        const INTERLACED = ffi::DRM_MODE_FB_INTERLACED;
+        /// Enables .modifier
+        const MODIFIERS = ffi::DRM_MODE_FB_MODIFIERS;
     }
 }
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -309,21 +309,29 @@ pub trait Device: super::Device {
     fn add_planar_framebuffer<B>(
         &self,
         planar_buffer: &B,
-        modifiers: &[Option<DrmModifier>; 4],
         flags: FbCmd2Flags,
     ) -> Result<framebuffer::Handle, SystemError>
     where
         B: buffer::PlanarBuffer + ?Sized,
     {
+        let modifier = planar_buffer.modifier();
+        let has_modifier = flags.contains(FbCmd2Flags::MODIFIERS);
+        assert!((has_modifier && modifier.is_some()) || (!has_modifier && modifier.is_none()));
+        let modifier = if let Some(modifier) = modifier {
+            u64::from(modifier)
+        } else {
+            0
+        };
+
         let (w, h) = planar_buffer.size();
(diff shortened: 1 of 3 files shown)


linkmauve:
Somewhat unrelated, but why is `buffer::PlanarBuffer` a trait and not just a struct?

Drakulix (changes requested):
Yes and no, reading through docs again, it looks like there is a way to specific no modifiers, which is not setting [`DRM_MODE_FB_MODIFIERS`](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/include/drm/drm_mode.h#L664) in `flags`.

The old code also got that wrong, but at least didn't force downstream to specify a non-sense modifier.

So this should take an `Option<DrmModifier>` and force `flags` to contain `DRM_MODE_FB_MODIFIERS`, if that `Option` is `Some`.

An even better solution would be to turn `flags` into a proper enum (and pass a `&[FramebufferFlags]` slice to this function. It would just contain [`interlaced`](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/include/drm/drm_mode.h#L663), so technically a bool would be enough, but that would mean new flags would force us to break the api in the future, while we could mark `FramebufferFlags` `non_exhaustive` today.

Drakulix:
> Somewhat unrelated, but why is `buffer::PlanarBuffer` a trait and not just a struct?

Because different types might contain drm-handles *and* accompanying information. E.g. dumb-buffer implements this trait in this library, but notably [gbm-rs](https://github.com/Smithay/gbm.rs) also contains an implementation of this trait for gbm `BufferObject`s.

linkmauve:
> Yes and no, reading through docs again, it looks like there is a way to specific no modifiers, which is not setting [`DRM_MODE_FB_MODIFIERS`](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/include/drm/drm_mode.h#L664) in `flags`.
> 
> The old code also got that wrong, but at least didn't force downstream to specify a non-sense modifier.

Note that not using modifiers (the old API) is the same as passing `INVALID` with the flag set.

> 
> So this should take an `Option<DrmModifier>` and force `flags` to contain `DRM_MODE_FB_MODIFIERS`, if that `Option` is `Some`.

So `Some(INVALID)` would be the same as None, that’s pretty much what I was suggesting.

> 
> An even better solution would be to turn `flags` into a proper enum (and pass a `&[FramebufferFlags]` slice to this function. It would just contain [`interlaced`](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/include/drm/drm_mode.h#L663), so technically a bool would be enough, but that would mean new flags would force us to break the api in the future, while we could mark `FramebufferFlags` `non_exhaustive` today.

I’ll have a look at that, thanks for the suggestion.

Drakulix:
> Note that not using modifiers (the old API) is the same as passing `INVALID` with the flag set. 
> So `Some(INVALID)` would be the same as None, that’s pretty much what I was suggesting.

Given that the docs [explicitly mention](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/include/drm/drm_fourcc.h#L481-484), that at least LINEAR isn't the same as the flag missing (though that is expected and very reasonable), I don't have high hopes, that some driver might not have different paths for `INVALID`+flag or no-modifier-flag.

I also don't see any IGT test specifically testing this: https://drm.pages.freedesktop.org/igt-gpu-tools/igt-kms-tests.html
So until proven otherwise, I would like the API to expose all argument combinations to the IOCTL, just to be safe.

I have seen enough weird bugs during compositor development, that I am not trusting this.

linkmauve:
Kernel devs told me the kernel doesn’t accept `INVALID` there actually, unlike the Wayland protocol for the same thing (`zwp_linux_dmabuf_v1`), so I was actually wrong.

I was wondering though, why isn’t the modifier part of the `PlanarBuffer` actually?  It is completely state that is part of the buffer, not of the call.

Drakulix:
> I was wondering though, why isn’t the modifier part of the PlanarBuffer actually? It is completely state that is part of the buffer, not of the call.

No good reason, it just has grown this way over time. Nothing about the PlanarBuffer requires the format to have a modifier, the original idea was just to model buffers with multiple planes.

That said, we could add a `modifier`-function just like `format` and make it return an `Option`.
But then you can also question why `Buffer` doesn't include `bpp`, that is required when creating a `DumbBuffer` anyway and can also be queried from a gbm-`BufferObject`.

linkmauve:
I’ve changed this implementation to respond to your comments.  One design decision I made was to keep the `DRM_MODE_FB_MODIFIERS` flag there, and assert if it was missing, but it could be better to rely on `Option` instead, let me know if you prefer that.

I’m not (much) interested in non-planar buffers, as this is legacy API at this point.

Drakulix (approved):
Works for me from an API perspective. Thanks for working on this.

src/control/mod.rs
```diff
@@ -1450,3 +1463,14 @@ bitflags::bitflags! {
...
     }
 }
+
+bitflags::bitflags! {
+    /// Planar framebuffer flags
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+    pub struct FbCmd2Flags : u32 {
+        /// For interlaced framebuffers
+        const INTERLACED = ffi::DRM_MODE_FB_INTERLACED;
+        /// Enables .modifier
```
MarijnS95, line 1473: Nit: would be great if this was an intradoc-link to ``[`PlanarInfo::modifier()`]`` and/or `PlanarBuffer::modifier()`?

src/control/mod.rs
```diff
@@ -304,21 +309,29 @@ pub trait Device: super::Device {
...
-        modifiers: &[Option<DrmModifier>; 4],
-        flags: u32,
+        flags: FbCmd2Flags,
     ) -> Result<framebuffer::Handle, SystemError>
     where
         B: buffer::PlanarBuffer + ?Sized,
     {
+        let modifier = planar_buffer.modifier();
+        let has_modifier = flags.contains(FbCmd2Flags::MODIFIERS);
+        assert!((has_modifier && modifier.is_some()) || (!has_modifier && modifier.is_none()));
```
MarijnS95, line 319: Nit: isn't this just `assert_eq!(has_modifier, modifier.is_some());`?

PR #163, opened 2023-09-11
Add functions for drm leasing

Drakulix:
Draft pull request until actually tested in smithay.

### DIFF
--- a/drm-ffi/src/mode.rs
+++ b/drm-ffi/src/mode.rs
@@ -780,6 +780,93 @@ pub fn atomic_commit(
     Ok(())
 }
 
+/// Create a drm lease
+pub fn create_lease(
+    fd: RawFd,
+    objects: &[u32],
+    flags: u32,
+) -> Result<drm_mode_create_lease, Error> {
+    let mut data = drm_mode_create_lease {
+        object_ids: objects.as_ptr() as _,
+        object_count: objects.len() as u32,
+        flags,
+        ..Default::default()
+    };
+
+    unsafe {
+        ioctl::mode::create_lease(fd, &mut data)?;
+    }
+
+    Ok(data)
+}
+
+/// List all active drm leases
+pub fn list_lessees(
+    fd: RawFd,
+    mut lessees: Option<&mut Vec<u32>>,
+) -> Result<drm_mode_list_lessees, Error> {
+    let mut sizes = drm_mode_list_lessees::default();
+
+    unsafe {
+        ioctl::mode::list_lessees(fd, &mut sizes)?;
+    };
+
+    map_reserve!(lessees, sizes.count_lessees as usize);
+
+    let mut data = drm_mode_list_lessees {
+        lessees_ptr: map_ptr!(&lessees),
+        count_lessees: map_len!(&lessees),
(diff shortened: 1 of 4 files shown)


drm-ffi/src/mode.rs
```diff
@@ -780,6 +780,93 @@ pub fn atomic_commit(
...
+    map_reserve!(lessees, sizes.count_lessees as usize);
+
+    let mut data = drm_mode_list_lessees {
+        lessees_ptr: map_ptr!(&lessees),
+        count_lessees: map_len!(&lessees),
+        ..Default::default()
+    };
+
+    unsafe {
+        ioctl::mode::list_lessees(fd, &mut data)?; // get count
```
i509VCB: not getting count here, you did that above.

Drakulix:
Tested in https://github.com/Smithay/smithay/pull/1132 with anvil and appears to work.

src/control/mod.rs
```diff
@@ -55,14 +55,19 @@ use std::iter::Zip;
...
 
 use core::num::NonZeroU32;
 
+pub use nix::fcntl::OFlag;
+
 /// Raw handle for a drm resource
 pub type RawResourceHandle = NonZeroU32;
 
+/// Id of a Lease
+pub type DrmLeaseId = NonZeroU32;
```
Slabity: Can probably just call this `LeaseId` to be consistent.

The only types we have prefixed in this crate with `Drm` is types from the `drm_fourcc` crate.

src/control/mod.rs
```diff
@@ -936,6 +941,35 @@ pub trait Device: super::Device {
         Ok(())
     }
 
+    /// Create a drm lease
+    fn create_lease(
+        &self,
+        objects: &[RawResourceHandle],
+        flags: OFlag,
+    ) -> Result<(DrmLeaseId, RawFd), SystemError> {
```
Slabity: Would separating the lease ID and underlying file descriptor be a good idea? Or wrapping them up into something that takes ownership to be dropped later/not outlive the underlying device?

Drakulix: Usually the file descriptor will be send to another process, while the lease_id will be kept in the current process to keep track of the lease and revoke it later.

So I guess we could technically wrap the `LeaseId`, so that it gets revoked on drop, but that needs access to the `control::Device` used to create, which potentially introduces lifetime issues. This would also be different compared to most other objects this crate crates, e.g. framebuffers, which are not cleaned up on drop.

But I guess we could wrap the `RawFd` in an `OwnedFd`, so it cannot be kept open accidentally, I will fix that.

Slabity: Ah, got it. The file descriptor gets sent to another process via UNIX socket or something? (I'm guessing an X11 or Wayland protocol?)

In that case, the `OwnedFd` is probably good.

Slabity:
I think this looks good. I'm not seeing any other issues so I'm going to merge this.

src/control/mod.rs
Drakulix: Yes exactly. Leasing basically lets you borrow a subset of drm resources into a new file descriptor without giving up DRM master.

PR #169, opened 2023-09-24
Reduce the number of get_property ioctls called

linkmauve:
I’m not super happy with the `as_hashmap()` name, feel free to bikeshed on that one. Especially because it does a lot of ioctls, that should get a name that looks more costly.

All together this reduces the number of `DRM_IOCTL_MODE_OBJ_GETPROPERTIES` from 32 to 12 in the atomic_modeset example, and the number of `DRM_IOCTL_MODE_GETPROPERTY` from 172 to 77.

A further improvement could be to only fetch the name (that is, do one call of getproperty instead of the whole call/allocation/call) and return a `HashMap<String, property::Handle>`, but non-example code probably want the whole property info to know what to actually set there.


Drakulix (approved):
Looks good to me. Thanks!

PR #171, opened 2023-10-04
Use `OwnedFd` and `BorrowedFd` where `RawFd` is used in API

ids1024:
This is a breaking API change.

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -56,7 +56,7 @@ use std::iter::Zip;
 use std::mem;
 use std::num::NonZeroUsize;
 use std::ops::RangeBounds;
-use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, OwnedFd, RawFd};
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, FromRawFd, OwnedFd, RawFd};
 use std::time::Duration;
 
 use core::num::NonZeroU32;
@@ -777,15 +777,19 @@ pub trait Device: super::Device {
     }
 
     /// Convert a prime file descriptor to a GEM buffer handle
-    fn prime_fd_to_buffer(&self, fd: RawFd) -> Result<buffer::Handle, SystemError> {
-        let info = ffi::gem::fd_to_handle(self.as_fd().as_raw_fd(), fd)?;
+    fn prime_fd_to_buffer(&self, fd: BorrowedFd<'_>) -> Result<buffer::Handle, SystemError> {
+        let info = ffi::gem::fd_to_handle(self.as_fd().as_raw_fd(), fd.as_raw_fd())?;
         Ok(from_u32(info.handle).unwrap())
     }
 
     /// Convert a GEM buffer handle to a prime file descriptor
-    fn buffer_to_prime_fd(&self, handle: buffer::Handle, flags: u32) -> Result<RawFd, SystemError> {
+    fn buffer_to_prime_fd(
+        &self,
+        handle: buffer::Handle,
+        flags: u32,
+    ) -> Result<OwnedFd, SystemError> {
         let info = ffi::gem::handle_to_fd(self.as_fd().as_raw_fd(), handle.into(), flags)?;
-        Ok(info.fd)
+        Ok(unsafe { OwnedFd::from_raw_fd(info.fd) })
     }
 
     /// Queue a page flip on the given crtc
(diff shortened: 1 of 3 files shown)


src/control/mod.rs
```diff
@@ -847,10 +851,11 @@ pub trait Device: super::Device {
     /// Imports a file descriptor exported by [`Self::syncobj_to_fd`] back into a process-local handle.
     fn fd_to_syncobj(
         &self,
-        fd: RawFd,
+        fd: BorrowedFd<'_>,
         import_sync_file: bool,
     ) -> Result<syncobj::Handle, SystemError> {
-        let info = ffi::syncobj::fd_to_handle(self.as_fd().as_raw_fd(), fd, import_sync_file)?;
+        let info =
+            ffi::syncobj::fd_to_handle(self.as_fd().as_raw_fd(), fd.as_raw_fd(), import_sync_file)?;
```
MarijnS95, line 858: Just double-checking: this function _does not_ consume ownership of the file descriptor? And it's the same for PRIME importing?

MarijnS95 (approved):
Together with this, does it makes sense to remove the `SyncFile` type and replace it with a simple `OwnedFd` instead of owning an `fs::File` inside and implementing the same common traits? That is closer to prime/buffers not having a `PrimeFile` type owning an `fd` either.

src/control/mod.rs
ids1024: I don't think an ioctl like this would close the file descriptor, but I'm having trouble finding any explicit documentation of this. I guess it shouldn't be too hard to test it to double check.

ids1024:
Yeah, if `SyncFile` does nothing other than wrap a `File` and implement `FromRawFd`/`AsFd`/`AsRawFd`, just using an `OwnedFd` probably makes more sense.

src/control/mod.rs
MarijnS95: Hopefully. The documentation doesn't mention but we might piece it together from the kernel source or a test indeed.

For example in Vulkan userspace drivers, file descriptor ownership is transferred to the Vulkan object (memory or fence) i.e. the caller _should not_ close the file descriptor after that.

ids1024: It makes sense that Vulkan would want to own the file descriptor, since it's a user-space library and probably needs to keep the fd to perform multiple ioctls with it.

MarijnS95: It is probably more convenient from an ownership-perspective? Or to automatically close the `fd` when the user destroys the imported fence or memory object. Fences are mostly single-use primitives anyway. If possible they can be `dup()`d when the kernel driver supports that sort of thing.

Unfortunately Windows handles don't transfer ownership, the caller is expected to call `CloseHandle()` instead: https://github.com/ash-rs/ash/pull/791. So it's all a bit vauge.

Drakulix: Was this checked, is this resolved?

ids1024: To be thorough, here's an example that tests this:

```rust
extern crate drm;
extern crate image;
extern crate nix;

use drm::buffer::Buffer;
use std::os::fd::{AsRawFd, AsFd, OwnedFd, RawFd};

pub mod utils;
use utils::*;

fn buffer_syncobj_fds() -> (OwnedFd, OwnedFd) {
    let card = Card::open_global();

    let buffer = card.create_dumb_buffer((1920, 1080), drm::buffer::DrmFourcc::Rgba8888, 8).unwrap();
    let buffer_fd = card.buffer_to_prime_fd(buffer.handle(), 0).unwrap();
    card.close_buffer(buffer.handle()).unwrap();

    let syncobj = card.create_syncobj(false).unwrap();
    let syncobj_fd = card.syncobj_to_fd(syncobj, false).unwrap();

    (buffer_fd, syncobj_fd)
}

fn fd_is_valid(fd: RawFd) -> bool {
    nix::fcntl::fcntl(fd, nix::fcntl::FcntlArg::F_GETFD).is_ok()
}

fn main() {
    let (buffer_fd, syncobj_fd) = buffer_syncobj_fds();
    let buffer_fd_raw = buffer_fd.as_raw_fd();
    let syncobj_fd_raw = syncobj_fd.as_raw_fd();

    let card = Card::open_global();


    dbg!(fd_is_valid(buffer_fd_raw));
    dbg!(card.prime_fd_to_buffer(buffer_fd.as_fd()).unwrap());
    dbg!(fd_is_valid(buffer_fd_raw));
    drop(buffer_fd);
    dbg!(fd_is_valid(buffer_fd_raw));

    dbg!(fd_is_valid(syncobj_fd_raw));
    dbg!(card.fd_to_syncobj(syncobj_fd.as_fd(), false).unwrap());
    dbg!(fd_is_valid(syncobj_fd_raw));
    drop(syncobj_fd);
    dbg!(fd_is_valid(syncobj_fd_raw));
}
```

Running (with a driver that supports syncobjs) reports that the fds remain valid after the calls to `prime_fd_to_buffer` and `fd_to_syncobj`. (And reports them as invalid after `drop` of the `OwnedFd` closes the fd.)

That seems satisfactory here. Though I didn't really expect a kernel-level API to do anything different here.

PR #172, opened 2023-10-08
drm-sys: add and update FreeBSD bindings

jbeich:
Generated inside 13.2 jail after ensuring nop on amd64 (had to downgrade libdrm). Originally for v0.9.0 (drm-sys 0.4.0) in order to support backports but later upgraded.

Note, FreeBSD can be deduped into architecture-neutral version.


ids1024:
This should be unnecessary now with https://github.com/Smithay/drm-rs/pull/176.

PR #174, opened 2023-10-18
drm-ffi: Accept `BorrowedFd` arguments instead of `RawFd`

ids1024:
This should make these functions safe, instead of technically violating IO safety rules as they currently do. (I can't say for sure there aren't other safety concerns with any of the ioctls, but this should be better anyway.)

This will be needed in `drm-ffi` switches to using `rustix` for ioctl calls.


Drakulix (approved):
LGTM

PR #175, opened 2023-10-24
Add riscv64 bindings

Vixea:


Drakulix:
I don't think there exists any easily accessible risc-based CI. So... I don't know if we want to include these bindings as a best effort (and make them go outdated frequently) or just advise users to use the `use_bindgen` feature...

Vixea:
Hmm, what would make these bindings get outdated I basically run my riscv machine constantly so if I could possibly find a way to check that it would be easy to solve

Drakulix:
> Hmm, what would make these bindings get outdated I basically run my riscv machine constantly so if I could possibly find a way to check that it would be easy to solve

New kernels

Vixea:
Considering this was run on kernel version 5.10 it might already be outdated

Drakulix:
Which is exactly why CI would be nice. Looking at our CI file, we already use QEMU + ubuntu-ports for a bunch of targets anyway and they seem to have a riscv64 target, we just don't generate for that, because nobody cared. Could you try to fix that?

Vixea:
I can try again, not exactly a ci master

ids1024:
> or just advise users to use the use_bindgen feature...

It's awkward when using Smithay to have to enable this for several different dependencies.

It could automatically use that when generated bindings don't already exist, but `bindgen` would have to be a hard dependency. Or `Cargo.toml` would have to specify it as required on exactly the architecture/OS pairs where there aren't currently bindings.

> New kernels

The kernel ABI should be stable, so the existing version continues to work, right? I guess it just needs to be regenerated when their are new ioctls/etc that drm-rs adds support for?

ids1024:
It seems the bindings here are the same as `x86_64` and `aarch64`. While `arm` and `x86` differ only in the definition of `__kernel_size_t`.

The same DRM headers are used on Linux and BSDs. See in https://github.com/torvalds/linux/blob/master/include/uapi/drm/drm.h the `#elif defined(__linux__)`, `#else /* One of the BSDs */`.

 So I think we could just generate one version of the bindings if we manually specify `__kernel_size_t` for Linux architectures where it varies. And `drm_handle_t`, which is `unsigned int` on Linux and `unsigned long` on BSD.

ids1024:
https://github.com/Smithay/drm-rs/pull/176 has a possible solution for this, though CI still needs to be updated.

Vixea:
Closed in favor of #176

PR #176, opened 2023-10-24
Generate bindings once for all platforms and architectures

ids1024:
The same drm headers are used for all architectures, and for the BSDs. There are only a couple typedefs that differ.

This uses `linux-raw-sys` to define `__kernel_size_t` on Linux, and `libc::size_t` on the BSDs.

I guess CI needs to be updated for this. Not sure if there would be a good way to automatically test that the bindings are correct for multiple architectures.


drm-ffi/drm-sys/src/lib.rs
```diff
@@ -3,15 +3,27 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 
+#[cfg(any(target_os = "android", target_os = "linux"))]
+extern crate linux_raw_sys;
+
+#[cfg(not(any(target_os = "android", target_os = "linux")))]
+extern crate libc;
```
MarijnS95: Is this still needed if we were to set edition 2018?

ids1024: No, the `extern` crate lines would be unnecessary in edition 2018 or later.

drm-ffi/drm-sys/build.rs
```diff
@@ -3,10 +3,30 @@ mod use_bindgen {
...
+    struct Callbacks;
+
+    impl ParseCallbacks for Callbacks {
+        fn blocklisted_type_implements_trait(
+            &self,
+            name: &str,
+            _derive_trait: DeriveTrait,
+        ) -> Option<ImplementsTrait> {
+            if name == "__kernel_size_t" || name == "drm_handle_t" {
+                Some(ImplementsTrait::Yes)
```
MarijnS95, line 22: Had to go to the [(imo rather vague) `bindgen` documentation](https://docs.rs/bindgen/latest/bindgen/callbacks/trait.ParseCallbacks.html#method.blocklisted_type_implements_trait) to figure out that this is telling `bindgen` that these types [implement a bunch of ("standard"??) traits](https://docs.rs/bindgen/latest/bindgen/callbacks/enum.DeriveTrait.html), so that any `struct` using these types could also automatically `derive()` those traits (all of them?).

Perhaps a doc-comment could complement this nicely?

drm-ffi/drm-sys/src/lib.rs
MarijnS95: I'm not sure what would limit us from setting it to `2018`, given that the MSRV is already at `1.65` and `drm-sys` can't be compiled with much older Rust anyway?

ids1024: Yeah, with the current MSRV I don't think there's any reason not to update to the current 2021 edition.

MarijnS95 (approved):
Really like the simplification this provides :ok_hand:

ids1024:
I guess if the drm headers are the same across operating systems and architectures, but have potential differences in the system headers they include and in an `#if` condition, this could be checked automatically by running bindgen for multiple platforms with the same drm header. Then it should find exactly identical results when we blocklist the parts that differ here.

But this should be correct currently, I doubt more differences will be added in the future, and they should be somewhat obvious looking at changes to `bindings.rs`.

Drakulix:
> I guess if the drm headers are the same across operating systems and architectures, but have potential differences in the system headers they include and in an `#if` condition, this could be checked automatically by running bindgen for multiple platforms with the same drm header. Then it should find exactly identical results when we blocklist the parts that differ here.
> 
> But this should be correct currently, I doubt more differences will be added in the future, and they should be somewhat obvious looking at changes to `bindings.rs`.

Right, CI should probably still use one architecture to generate the bindings (e.g. x86_64, given that is the least overhead) and then run the other architectures just to compare and fail the test if they differ.

Like that the crate can theoretically support anything, but we still would have a set of well tested architectures. 

Given #175 we probably would also want to add testing for riscv in this PR, no?

ids1024:
It's possible to generate bindings for headers on a specific architecture with something like:

```
apt-get install --yes wget linux-libc-dev-riscv64-cross bindgen libclang-dev rustfmt
wget https://github.com/torvalds/linux/raw/master/include/uapi/drm/drm.h
wget https://github.com/torvalds/linux/raw/master/include/uapi/drm/drm_mode.h
bindgen drm.h -- -D __user= -isysroot /usr/riscv64-linux-gnu
```

(Though that particular `bindgen` invocation doesn't generate exactly what we want.)

Then we can generate up-to-date bindings easily on CI, and verify that there aren't platform differences for any of the architectures Debian has compilers for.

Does it still make sense to have a `use_bindgen` feature flag if the pre-generated bindings work across platforms?

Drakulix:
> Does it still make sense to have a use_bindgen feature flag if the pre-generated bindings work across platforms?

Depends on how up-to-date our bindings are. I guess it doesn't hurt to keep it around?

ids1024:
Updated with CI changes to pull headers from Linux git repo, generate on all tested architectures, and compare to verify they are the same.

Drakulix (approved):
Amazing. Thanks a bunch!

PR #180, opened 2023-11-07
Use Rustix

ids1024:
Updates `drm-ffi` to use Rustix.

Based on https://github.com/Smithay/drm-rs/pull/179.
    
Not *too* hard to update by defining macros like Nix had, though it's probably not ideal.

The generic ioctl API for Rustix seems a bit awkward when dealing with APIs like drm that involve very large numbers of ioctls. Instead of generating functions with a macro like this, it's possible to define types aliases with `ReadWriteOpcode`, etc. But it's still necessary to deal with `Getter`/`Setter`, etc...


ids1024:
With https://github.com/bytecodealliance/rustix/pull/925 and https://github.com/bytecodealliance/rustix/pull/926 merged and after a few more changes here, this should be good to merged.

Drakulix (approved):
Thanks!

PR #183, opened 2023-12-01
0.38.22 Has Been Yanked

njwhite:
See https://crates.io/crates/rustix/versions !


ids1024:
`version = "0.38.22` means this depends on 0.38.22 or later. Just `version = 0.38` would be less correct since it wouldn't build with `0.38.0` and needs newer features.

This could be bumped to the latest patch release, but I don't think that's necessary. Cargo won't use the yanked version unless it's already in the lock file, and `version = 0.38` wouldn't be any different, if the lockfile already contains `0.38.22`.

PR #184, opened 2023-12-04
drm-sys: Update `bindgen`

ids1024:
This doesn't have any impact on the generated bindings.


Drakulix:
Any reason then to upgrade? Couldn't this for example raise the MSRV?

ids1024:
Just so that `cargo outdated` doesn't list it. Pretty much. And to see if it makes a difference in the bindings (which it didn't).

Bindgen has `rust-version = "1.60.0"`, so there shouldn't be any problem with it. (Not that most users of the crate will be using the `use_bindgen` feature anyway).

PR #189, opened 2024-01-19
ffi: fix build with libdrm 2.4.120

davide125:
libdrm 2.4.120 introduced a new field with https://gitlab.freedesktop.org/mesa/drm/-/commit/6c4392f49bdd0b34590ae646868229318f880f81


davide125:
Ah, of course this breaks with older versions. We might need to make this conditional somehow.

PolyMeilex:
I belive this can be closed now, https://github.com/Smithay/drm-rs/pull/190  added missing `deadline_nsec` fields

Issue #186, opened 2023-12-06
Async atomic flips


i509VCB:
https://patchwork.freedesktop.org/series/120103/ has been merged from what I heard over on wlroots.

PR #190, opened 2024-02-17
Update drm sys to get new caps

PolyMeilex:
Atomic async page flip needs new capabilities from the kernel, this updates drm-rs to include those, as well as some unrelated Client caps.

related: https://github.com/Smithay/smithay/pull/1325


PolyMeilex:
Well this fails as CI generates its own binding for the older version of DRM.
Ideas on how to tackle this welcome.

i509VCB:
Fixes #186  as well

PolyMeilex:
closes #189 as well

Drakulix:
> Well this fails as CI generates its own binding for the older version of DRM. Ideas on how to tackle this welcome.

The best solution would be to query the headers from another source. E.g. directly from kernel sources.

PolyMeilex:
Done, included the kernel headers and added a feature to generate from them instead of libdrm ones.

Now it looks like latest nightly rust is not happy about our type casts, that appear to be correct as far as I can tell:
<details>
<summary>Error logs</summary>
<code>
error: casting references to a bigger memory layout than the backing allocation is undefined behavior, even if the reference is unused
    --> src/control/mod.rs:1133:34
     |
1128 |             let event = unsafe { &*(self.event_buf.as_ptr().add(self.i) as *const ffi::drm_event) };
     |                                   --------------------------------------------------------------- backing allocation comes from here
...
1133 |                         unsafe { &*(event as *const _ as *const ffi::drm_event_vblank) };
     |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: casting from `drm_event` (8 bytes) to `drm_event_vblank` (32 bytes)
     = note: `#[deny(invalid_reference_casting)]` on by default

error: casting references to a bigger memory layout than the backing allocation is undefined behavior, even if the reference is unused
    --> src/control/mod.rs:1147:34
     |
1128 |             let event = unsafe { &*(self.event_buf.as_ptr().add(self.i) as *const ffi::drm_event) };
     |                                   --------------------------------------------------------------- backing allocation comes from here
...
1147 |                         unsafe { &*(event as *const _ as *const ffi::drm_event_vblank) };
     |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: casting from `drm_event` (8 bytes) to `drm_event_vblank` (32 bytes)
</code>
</details>

Anyway, that's unrelated to this PR. Disabled the lint for now

PolyMeilex:
Ping, anything more I can do to move this forward?

Drakulix:
> Ping, anything more I can do to move this forward?

Sorry, forgot to take a look at this. Looks mostly good to me, I wonder though, if we can add the update.sh as a step to the `update-bindings`/`compare-bindings` ci job, so that we always compare with up-to-date header files and perhaps also include updated header files in the automatically created pull requests?

PolyMeilex:
> > Ping, anything more I can do to move this forward?
> 
> Sorry, forgot to take a look at this. Looks mostly good to me, I wonder though, if we can add the update.sh as a step to the `update-bindings`/`compare-bindings` ci job, so that we always compare with up-to-date header files and perhaps also include updated header files in the automatically created pull requests?

Added a download headers step right before bindgen is called, and this seems to do the trick (at least on my test repo)

Drakulix:
I am confused, looking through your PR, I saw that we already added downloading kernel headers at some point here: https://github.com/Smithay/drm-rs/pull/176/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR195-R199

So maybe that isn't necessary and we can just merge this without?

PolyMeilex:
Dropped the kernel gen commits and created the [backup branch for them](https://github.com/PolyMeilex/drm-rs/tree/update-drm-sys-to-get-new-caps-backup)

But yeah, the CI seems to be downgrading the bindings now :shrug:

PolyMeilex:
The `Fetch drm headers` seems to be only called for `test` step, so that's probably why checks fail.

So do we disable regeneration for check steps, or fetch in them as well?

Drakulix:
> So do we disable regeneration for check steps, or fetch in them as well?

I guess fetch as well? We would want to use the same ones throughout after all.

PolyMeilex:
@Drakulix Done, we now fetch on check steps as well, and CI is happy

Drakulix (approved):
Thanks!

PR #191, opened 2024-03-19
Fix unaligned memory access when reading DRM events

tronical:
Running with debug assertions enabled on armv7 (an stm32mp157) panics due to unaligned memory reads:

thread 'main' panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0xbec90b44', .../drm-0.9.0/src/control/mod.rs:906:34

Fix this by using the corresponding functions from core to safely copy the data (in C we'd do a memcpy).

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -1139,14 +1139,16 @@ impl Iterator for Events {
 
     fn next(&mut self) -> Option<Event> {
         if self.amount > 0 && self.i < self.amount {
-            let event = unsafe { &*(self.event_buf.as_ptr().add(self.i) as *const ffi::drm_event) };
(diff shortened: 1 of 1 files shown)


Drakulix:
Could you rebase this one as well please? :)

tronical:
Done. Sorry, took a while :). The `#[allow(unknown_lints)]  #[allow(invalid_reference_casting)]` is still there, as I'm not sure if they can be removed. How can I test?

Drakulix:
> Done. Sorry, took a while :). The `#[allow(unknown_lints)] #[allow(invalid_reference_casting)]` is still there, as I'm not sure if they can be removed. How can I test?

Just remove them and see if CI fails. If it does we need to add them again. (I think the error comes from different type-aliases on different platforms, so this is difficult to test outside of CI.)

Drakulix:
Thanks!

PR #192, opened 2024-03-20
Fix drm::control::Device::add_planar_buffer when GBM Bo uses DrmModifier::Invalid

tronical:
When adding a GBM supplied buffer via add_planar_buffer, since commit 6cb3a27d5b3f4fecb5415bed69497491759126a9 the modifiers are fetched directly from the planar buffer trait. What the code does not take into account though is when the GBM bo returns DrmModifier::Invalid. In that case add_fb2 expects 0 as modifier in the array for each plane, instead of DRM_FORMAT_MOD_INVALID.

### DIFF
--- a/src/control/mod.rs
+++ b/src/control/mod.rs
@@ -353,7 +353,9 @@ pub trait Device: super::Device {
     where
         B: buffer::PlanarBuffer + ?Sized,
     {
-        let modifier = planar_buffer.modifier();
+        let modifier = planar_buffer
+            .modifier()
+            .filter(|modifier| !matches!(modifier, DrmModifier::Invalid));
         let has_modifier = flags.contains(FbCmd2Flags::MODIFIERS);
         assert!((has_modifier && modifier.is_some()) || (!has_modifier && modifier.is_none()));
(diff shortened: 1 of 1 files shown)


tronical:
One could argue that the `PlanarBuffer` implementation of `BufferObject` in gbm, `fn modifiers(&self) -> Option<Modifier>` should return `None` for DrmModifier::Invalid. I can make a PR for that - separately - but meanwhile I think since the `::Invalid` variant _does_ exist, it makes sense to "support" it.

Drakulix (approved):

tronical:
I force-pushed to get the formatting fixed (my bad).

tronical:
The CI failure looks bogus. The error about the invalid cast looks like it was from a build of an earlier sha. The latest build in https://github.com/Smithay/drm-rs/actions/runs/8346623741 has only the (seemingly unrelated) drm_syncobj failures.

Drakulix:
Could you rebase? https://github.com/Smithay/drm-rs/pull/190 should have fixed CI.

tronical:
Done :)

Drakulix:
Thanks!

PR #201, opened 2024-08-14
Add `(major, minor, patchlevel)` to `Driver` `get_version()` probe

MarijnS95:
`struct drm_version` contains actual version fields, beyond the strings for the driver name, deveice description and date string.  Add a tuple field for this to `struct Driver`.


Drakulix (approved):
thx!

PR #202, opened 2024-09-06
Add DRM node abstraction

chrisduerr:
This copies the DRM node abstraction from Smithay almost 1:1, to make it easier for other consumers to reuse this code.

### DIFF
--- a/drm-ffi/Cargo.toml
+++ b/drm-ffi/Cargo.toml
@@ -5,7 +5,7 @@ repository = "https://github.com/Smithay/drm-rs"
 version = "0.8.0"
 license = "MIT"
 authors = ["Tyler Slabinski <tslabinski@slabity.net>"]
-rust-version = "1.65"
+rust-version = "1.66"
 edition = "2021"
 
 [dependencies]
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -3,6 +3,7 @@
 pub mod constants;
 
 use std::error::Error;
+use std::fmt::Debug;
 use std::fmt::{self, Display, Formatter};
 use std::io;
 use std::os::unix::io::AsFd;
@@ -156,15 +157,7 @@ impl NodeType {
 
 impl Display for NodeType {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "{}",
-            match self {
-                NodeType::Primary => "Primary",
-                NodeType::Control => "Control",
-                NodeType::Render => "Render",
-            }
-        )
+        Debug::fmt(self, f)
     }
 }
 
@@ -181,7 +174,7 @@ pub enum CreateDrmNodeError {
 impl Display for CreateDrmNodeError {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
-            Self::Io(err) => err.fmt(f),
+            Self::Io(err) => Display::fmt(err, f),
             Self::NotDrmNode => write!(
                 f,
                 "the provided file descriptor does not refer to a DRM node"
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -175,10 +175,9 @@ impl Display for CreateDrmNodeError {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
             Self::Io(err) => Display::fmt(err, f),
-            Self::NotDrmNode => write!(
-                f,
-                "the provided file descriptor does not refer to a DRM node"
-            ),
+            Self::NotDrmNode => {
+                f.write_str("the provided file descriptor does not refer to a DRM node")
+            }
         }
     }
 }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -286,7 +286,7 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         // Only 1 primary, control and render node may exist simultaneously, so the
         // first occurrence is good enough.
         if name.starts_with(ty.minor_name_prefix()) {
-            let path = [r"/", "dev", "dri", &name].iter().collect::<PathBuf>();
+            let path = Path::new("/dev/dri").join(&*name);
             if path.exists() {
                 return Ok(path);
             }
@@ -324,7 +324,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         if let Ok(old_id) = suffix.parse::<u32>() {
             let id_mask = 0b11_1111;
             let id = old_id & id_mask + get_minor_base(ty);
-            let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
+            let path = PathBuf::new("/dev/dri").join(ty.minor_name_prefix()).join(id);
             if path.exists() {
                 return Ok(path);
             }
@@ -356,7 +356,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     let old_id = minor(dev);
     let id_mask = 0b11_1111;
     let id = old_id & id_mask + get_minor_base(ty);
-    let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
+    let path = PathBuf::from("/dev/dri", ty.minor_name_prefix(), id);
     if path.exists() {
         return Ok(path);
     }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -333,9 +333,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         if let Ok(old_id) = suffix.parse::<u32>() {
             let id_mask = 0b11_1111;
             let id = old_id & id_mask + ty.minor_base();
-            let path = PathBuf::new("/dev/dri")
-                .join(ty.minor_name_prefix())
-                .join(id);
+            let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
             if path.exists() {
                 return Ok(path);
             }
@@ -368,7 +366,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     let old_id = minor(dev);
     let id_mask = 0b11_1111;
     let id = old_id & id_mask + ty.minor_base();
-    let path = PathBuf::from("/dev/dri", ty.minor_name_prefix(), id);
+    let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
     if path.exists() {
         return Ok(path);
     }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -3,8 +3,7 @@
 pub mod constants;
 
 use std::error::Error;
-use std::fmt::Debug;
-use std::fmt::{self, Display, Formatter};
+use std::fmt::{self, Debug, Display, Formatter};
 use std::io;
 use std::os::unix::io::AsFd;
 use std::path::{Path, PathBuf};
@@ -168,7 +167,7 @@ impl Display for NodeType {
     }
 }
 
-/// An error that may occur when creating a DrmNode from a file descriptor.
+/// An error that may occur when creating a [`DrmNode`] from a file descriptor.
 #[derive(Debug)]
 pub enum CreateDrmNodeError {
     /// Some underlying IO error occured while trying to create a DRM node.
@@ -232,8 +231,7 @@ fn devname(dev: dev_t) -> Option<String> {
     Some(String::from_utf8(dev_name).expect("Returned device name is not valid utf8"))
 }
 
-/// Returns if the given device by major:minor pair is a drm device
-///
+/// Returns if the given device by major:minor pair is a DRM device.
 #[cfg(target_os = "linux")]
 pub fn is_device_drm(dev: dev_t) -> bool {
     // We `stat` the path rather than comparing the major to support dynamic device numbers:
@@ -242,7 +240,7 @@ pub fn is_device_drm(dev: dev_t) -> bool {
     stat(path.as_str()).is_ok()
 }
 
-/// Returns if the given device by major:minor pair is a drm device
+/// Returns if the given device by major:minor pair is a DRM device.
 #[cfg(target_os = "freebsd")]
 pub fn is_device_drm(dev: dev_t) -> bool {
     devname(dev).map_or(false, |dev_name| {
@@ -253,7 +251,7 @@ pub fn is_device_drm(dev: dev_t) -> bool {
     })
 }
 
-/// Returns if the given device by major:minor pair is a drm device
+/// Returns if the given device by major:minor pair is a DRM device.
 #[cfg(not(any(target_os = "linux", target_os = "freebsd")))]
 pub fn is_device_drm(dev: dev_t) -> bool {
     major(dev) == DRM_MAJOR
@@ -375,7 +373,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     Err(io::Error::new(
         ErrorKind::NotFound,
         format!(
-            "Could not find node of type {} from DRM device {}:{}",
+            "Could not find node of type {} for DRM device {}:{}",
             ty,
             major(dev),
             minor(dev)
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -323,7 +323,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         let suffix = dev_name.trim_start_matches(|c: char| !c.is_numeric());
         if let Ok(old_id) = suffix.parse::<u32>() {
             let id_mask = 0b11_1111;
-            let id = old_id & id_mask;
+            let id = old_id & id_mask + get_minor_base(ty);
             let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
             if path.exists() {
                 return Ok(path);
@@ -355,7 +355,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
 
     let old_id = minor(dev);
     let id_mask = 0b11_1111;
-    let id = old_id & id_mask;
+    let id = old_id & id_mask + get_minor_base(ty);
     let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
     if path.exists() {
         return Ok(path);
@@ -370,4 +370,13 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
             minor(dev)
         ),
     ))
+}
+
+#[cfg(any(target_os = "freebsd", target_os = "openbsd"))]
+fn get_minor_base(type_: NodeType) -> u32 {
+    match type_ {
+        NodeType::Primary => 0,
+        NodeType::Control => 64,
+        NodeType::Render => 128,
+    }
 }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -233,8 +233,11 @@ fn devname(dev: dev_t) -> Option<String> {
 }
 
 /// Returns if the given device by major:minor pair is a drm device
+///
 #[cfg(target_os = "linux")]
 pub fn is_device_drm(dev: dev_t) -> bool {
+    // We `stat` the path rather than comparing the major to support dynamic device numbers:
+    //   https://gitlab.freedesktop.org/mesa/drm/-/commit/f8392583418aef5e27bfed9989aeb601e20cc96d
     let path = format!("/sys/dev/char/{}:{}/device/drm", major(dev), minor(dev));
     stat(path.as_str()).is_ok()
 }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -22,8 +22,6 @@ pub struct DrmNode {
 
 impl DrmNode {
     /// Creates a DRM node from an open drm device.
-    ///
-    /// This function does not take ownership of the passed in file descriptor.
     pub fn from_file<A: AsFd>(file: A) -> Result<DrmNode, CreateDrmNodeError> {
         let stat = fstat(file).map_err(Into::<io::Error>::into)?;
         DrmNode::from_stat(stat)
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -255,7 +255,7 @@ pub fn path_to_type<P: AsRef<Path>>(path: P, ty: NodeType) -> io::Result<PathBuf
     dev_path(stat.st_rdev, ty)
 }
 
-/// Returns the path of a specific type of node from the same DRM device as an existing DrmNode.
+/// Returns the path of a specific type of node from the same DRM device as an existing [`DrmNode`].
 pub fn node_path(node: &DrmNode, ty: NodeType) -> io::Result<PathBuf> {
     dev_path(node.dev, ty)
 }
@@ -322,18 +322,8 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     if let Some(dev_name) = devname(dev) {
         let suffix = dev_name.trim_start_matches(|c: char| !c.is_numeric());
         if let Ok(old_id) = suffix.parse::<u32>() {
-            let old_ty = match old_id >> 6 {
-                0 => NodeType::Primary,
-                1 => NodeType::Control,
-                2 => NodeType::Render,
-                _ => {
-                    return Err(io::Error::new(
-                        ErrorKind::NotFound,
-                        format!("{}:{} is no DRM device", major(dev), minor(dev)),
-                    ));
-                }
-            };
-            let id = old_id - get_minor_base(old_ty) + get_minor_base(ty);
+            let id_mask = 0b11_1111;
+            let id = old_id & id_mask;
             let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
             if path.exists() {
                 return Ok(path);
@@ -364,18 +354,8 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     }
 
     let old_id = minor(dev);
-    let old_ty = match old_id >> 6 {
-        0 => NodeType::Primary,
-        1 => NodeType::Control,
-        2 => NodeType::Render,
-        _ => {
-            return Err(io::Error::new(
-                ErrorKind::NotFound,
-                format!("{}:{} is no DRM device", major(dev), minor(dev)),
-            ));
-        }
-    };
-    let id = old_id - get_minor_base(old_ty) + get_minor_base(ty);
+    let id_mask = 0b11_1111;
+    let id = old_id & id_mask;
     let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
     if path.exists() {
         return Ok(path);
@@ -390,13 +370,4 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
             minor(dev)
         ),
     ))
-}
-
-#[cfg(any(target_os = "freebsd", target_os = "openbsd"))]
-fn get_minor_base(type_: NodeType) -> u32 {
-    match type_ {
-        NodeType::Primary => 0,
-        NodeType::Control => 64,
-        NodeType::Render => 128,
-    }
 }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -151,7 +151,7 @@ impl NodeType {
         }
     }
 
-    #[cfg(any(target_os = "freebsd", target_os = "openbsd"))]
+    #[cfg(not(target_os = "linux"))]
     fn minor_base(&self) -> u32 {
         match self {
             NodeType::Primary => 0,
@@ -351,7 +351,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
 }
 
 /// Returns the path of a specific type of node from the DRM device described by major and minor device numbers.
-#[cfg(target_os = "openbsd")]
+#[cfg(not(any(target_os = "linux", target_os = "freebsd")))]
 fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     use std::io::ErrorKind;
 
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -42,7 +42,7 @@ impl DrmNode {
     }
 
     /// Creates a DRM node from a [`dev_t`].
-    pub fn from_dev_id(dev: dev_t) -> Result<DrmNode, CreateDrmNodeError> {
+    pub fn from_dev_id(dev: dev_t) -> Result<Self, CreateDrmNodeError> {
         if !is_device_drm(dev) {
             return Err(CreateDrmNodeError::NotDrmNode);
         }
@@ -151,6 +151,15 @@ impl NodeType {
             NodeType::Primary => PRIMARY_NAME,
             NodeType::Control => CONTROL_NAME,
             NodeType::Render => RENDER_NAME,
+        }
+    }
+
+    #[cfg(any(target_os = "freebsd", target_os = "openbsd"))]
+    fn minor_base(&self) -> u32 {
+        match self {
+            NodeType::Primary => 0,
+            NodeType::Control => 64,
+            NodeType::Render => 128,
         }
     }
 }
@@ -323,8 +332,10 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         let suffix = dev_name.trim_start_matches(|c: char| !c.is_numeric());
         if let Ok(old_id) = suffix.parse::<u32>() {
             let id_mask = 0b11_1111;
-            let id = old_id & id_mask + get_minor_base(ty);
-            let path = PathBuf::new("/dev/dri").join(ty.minor_name_prefix()).join(id);
+            let id = old_id & id_mask + ty.minor_base();
+            let path = PathBuf::new("/dev/dri")
+                .join(ty.minor_name_prefix())
+                .join(id);
             if path.exists() {
                 return Ok(path);
             }
@@ -342,6 +353,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     ))
 }
 
+/// Returns the path of a specific type of node from the DRM device described by major and minor device numbers.
 #[cfg(target_os = "openbsd")]
 fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     use std::io::ErrorKind;
@@ -355,7 +367,7 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
 
     let old_id = minor(dev);
     let id_mask = 0b11_1111;
-    let id = old_id & id_mask + get_minor_base(ty);
+    let id = old_id & id_mask + ty.minor_base();
     let path = PathBuf::from("/dev/dri", ty.minor_name_prefix(), id);
     if path.exists() {
         return Ok(path);
@@ -370,13 +382,4 @@ fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
             minor(dev)
         ),
     ))
-}
-
-#[cfg(any(target_os = "freebsd", target_os = "openbsd"))]
-fn get_minor_base(type_: NodeType) -> u32 {
-    match type_ {
-        NodeType::Primary => 0,
-        NodeType::Control => 64,
-        NodeType::Render => 128,
-    }
 }
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -145,7 +145,7 @@ impl NodeType {
     /// Returns a string representing the prefix of a minor device's name.
     ///
     /// For example, on Linux with a primary node, the returned string would be `card`.
-    pub fn minor_name_prefix(&self) -> &str {
+    pub fn minor_name_prefix(&self) -> &'static str {
         match self {
             NodeType::Primary => PRIMARY_NAME,
             NodeType::Control => CONTROL_NAME,
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -198,13 +198,6 @@ impl From<io::Error> for CreateDrmNodeError {
     }
 }
 
-/// Returns if the given device by major:minor pair is a drm device
-#[cfg(target_os = "linux")]
-pub fn is_device_drm(dev: dev_t) -> bool {
-    let path = format!("/sys/dev/char/{}:{}/device/drm", major(dev), minor(dev));
-    stat(path.as_str()).is_ok()
-}
-
 #[cfg(target_os = "freebsd")]
 fn devname(dev: dev_t) -> Option<String> {
     use std::os::raw::{c_char, c_int};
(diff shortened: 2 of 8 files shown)


src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct DrmNode {
+    dev: dev_t,
+    ty: NodeType,
+}
+
+impl DrmNode {
+    /// Creates a DRM node from an open drm device.
+    ///
+    /// This function does not take ownership of the passed in file descriptor.
```
MarijnS95: Doesn't `AsFd` already signify that? (Though one could both move or borrow an `OwnedFd` with this API, IIRC)

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+        write!(f, "{}{}", self.ty.minor_name_prefix(), minor(self.dev_id()))
+    }
+}
+
+/// A type of node
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub enum NodeType {
+    /// A primary node may be used to allocate buffers.
+    ///
+    /// If no other node is present, this may be used to post a buffer to an output with mode-setting.
```
MarijnS95, line 128: I've never fully understood this, but does this ultimately decide whether we can derive `control::Device` in addition to `Device`?

And as with my `queue_/get_sequence() PR https://github.com/Smithay/drm-rs/pull/200, do some things currently in `control` belong to the "primary" device?

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    ///
+    /// Mode-setting is not possible with a render node.
+    Render,
+}
+
+impl NodeType {
+    /// Returns a string representing the prefix of a minor device's name.
+    ///
+    /// For example, on Linux with a primary node, the returned string would be `card`.
+    pub fn minor_name_prefix(&self) -> &str {
```
MarijnS95:
```suggestion
    pub fn minor_name_prefix(&self) -> &'static str {
```
 Perhaps?

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                NodeType::Primary => "Primary",
+                NodeType::Control => "Control",
+                NodeType::Render => "Render",
+            }
+        )
```
MarijnS95: This looks like the `Debug` implementation?

Perhaps use `f.write_str()` directly?

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+/// Returns if the given device by major:minor pair is a drm device
+#[cfg(target_os = "freebsd")]
+pub fn is_device_drm(dev: dev_t) -> bool {
+    devname(dev).map_or(false, |dev_name| {
+        dev_name.starts_with("drm/")
+            || dev_name.starts_with("dri/card")
+            || dev_name.starts_with("dri/control")
+            || dev_name.starts_with("dri/renderD")
+    })
+}
```
MarijnS95, line 252: Couldn't you compare `major()` against the constant here?

EDIT: Looks like you do that for non-feebsd non-linux. I thought it would be valid for at least Linux?

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// The provided file descriptor does not refer to a DRM node.
+    NotDrmNode,
+}
+
+impl Display for CreateDrmNodeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Io(err) => err.fmt(f),
+            Self::NotDrmNode => write!(
+                f,
```
MarijnS95:
```suggestion
            Self::NotDrmNode => f.write_str(
```

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    major(dev) == DRM_MAJOR
+}
+
+/// Returns the path of a specific type of node from the same DRM device as another path of the same node.
+pub fn path_to_type<P: AsRef<Path>>(path: P, ty: NodeType) -> io::Result<PathBuf> {
+    let stat = stat(path.as_ref()).map_err(Into::<io::Error>::into)?;
+    dev_path(stat.st_rdev, ty)
+}
+
+/// Returns the path of a specific type of node from the same DRM device as an existing DrmNode.
```
MarijnS95: As a path, right? Or is this referring to the DRM device _from an_ existing DrmNode?

(Intradoc link, too)

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+        1 => NodeType::Control,
+        2 => NodeType::Render,
+        _ => {
+            return Err(io::Error::new(
+                ErrorKind::NotFound,
+                format!("{}:{} is no DRM device", major(dev), minor(dev)),
+            ));
+        }
+    };
+    let id = old_id - get_minor_base(old_ty) + get_minor_base(ty);
```
MarijnS95: How about masking and shifting to not have to define this extra function which essentially returns a constant left-shifted by 6?

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
```
MarijnS95, line 368: (here and elsewhere) instead of the name prefix, couldn't we open `/dev/char/{DRM_MAJOR}:{id}` which this is typically a symlink to?

src/node/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    ))?;
+
+    for entry in read.flatten() {
+        let name = entry.file_name();
+        let name = name.to_string_lossy();
+
+        // Only 1 primary, control and render node may exist simultaneously, so the
+        // first occurrence is good enough.
+        if name.starts_with(ty.minor_name_prefix()) {
+            let path = [r"/", "dev", "dri", &name].iter().collect::<PathBuf>();
```
MarijnS95: Same here, I expected this to use the generic function (with masking and shifting). Any comment here (and in  `is_device_drm()`) why these are special-cased for Linux and FreeBSD?

MarijnS95 (commented):
Nice stuff! Something like this would also be great to start using in the examples, as we hardcode `card0` which even on Linux don't fare well when a GPU probes as `card1`.

One thing I don't follow is why - at least for Linux, I don't know anything about FreeBSD - certain implementations are special-cased. For at least Linux I believe the common implementation should be valid?

src/node/mod.rs
chrisduerr: I don't see much harm in clarifying either way, but generally you're correct.

src/node/mod.rs
chrisduerr: Good catch, no reason not to.

src/node/mod.rs
chrisduerr: I'd have to delegate this question to @i509VCB. I'm not familiar with these traits.

src/node/mod.rs
chrisduerr: I can see arguments going both ways tbh, this way the consistency of the implementation would be ensured at least.

But I've changed it to depend on the debug impl instead.

src/node/mod.rs
chrisduerr: I would assume it's done this way for a reason, but I'm not familiar with freebsd internals.

src/node/mod.rs
chrisduerr: From an existing DrmNode, but of a different node type.

src/node/mod.rs
chrisduerr: The paths are different for freebsd and linux, so why wouldn't it be special-cased? A difference here is to be expected, no?

src/node/mod.rs
chrisduerr: I've switched it to a bit mask instead.

src/node/mod.rs
chrisduerr: I don't see why we would? Accessing it directly seems superior to accessing a symlink? So why force ourselves to read a symlink when we can just construct the real path directly?

src/node/mod.rs
MarijnS95: Sorry, I also wanted to ask this question for Linux (lots to type on a small phone screen), which is also special-cased. Doesn't this work on Linux?

```rust
#[cfg(not(any(target_os = "linux", target_os = "freebsd")))]
pub fn is_device_drm(dev: dev_t) -> bool {
    major(dev) == DRM_MAJOR
}
```

src/node/mod.rs
MarijnS95: You're right, I misremembered and mis-suggested that `/dev/dri/cardX` is a symlink to `/dev/char/{DRM_MAJOR}:{id}` - it is the other way around.

src/node/mod.rs
MarijnS95: But is there a difference between Linux and not(Linux, FreeBSD)?

src/node/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+            "Could not find node of type {} from DRM device {}:{}",
+            ty,
+            major(dev),
+            minor(dev)
+        ),
+    ))
+}
+
+#[cfg(target_os = "openbsd")]
+fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
```
MarijnS95: Looks like the `cfg` got accidentally duplicated with the one above, this was `not()` before right?

src/node/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+        return Err(io::Error::new(
+            ErrorKind::NotFound,
+            format!("{}:{} is no DRM device", major(dev), minor(dev)),
+        ));
+    }
+
+    let old_id = minor(dev);
+    let id_mask = 0b11_1111;
+    let id = old_id & id_mask;
+    let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
```
MarijnS95: Shouldn't this OR the type back in? If I have a device with ID `1`, and I want to get a render device, I prefix with `renderD` and OR `2<<6` back into it, right?

src/node/mod.rs
MarijnS95: I see something weird on my machine which explains why you might not want to use IDs and bitmasks, but this logic checking inside `device/drm` instead:

```console
❯ l /dev/dri/by-path/
total 0
drwxr-xr-x 2 root root 120 Sep  6 15:00 ./
drwxr-xr-x 3 root root 140 Sep  6 15:00 ../
lrwxrwxrwx 1 root root   8 Sep  6 15:00 pci-0000:23:00.0-card -> ../card1
lrwxrwxrwx 1 root root  13 Sep  6 15:00 pci-0000:23:00.0-render -> ../renderD128
lrwxrwxrwx 1 root root   8 Sep  6 15:00 pci-0000:4c:00.0-card -> ../card2
lrwxrwxrwx 1 root root  13 Sep  6 15:00 pci-0000:4c:00.0-render -> ../renderD129
```

```console
❯ stat /dev/dri/* --format "%n: %R"
/dev/dri/by-path: 0
/dev/dri/card1: e201
/dev/dri/card2: e202
/dev/dri/renderD128: e280
/dev/dri/renderD129: e281
```

As you can see primary device `card1` has number `0` for the render device (if we mask the type bits out of it). Same for `card2` having number `1` in `renderD129`.

src/node/mod.rs
MarijnS95: For completeness it does seem that the `drm/` folder inside a device always contains a control device that points to the primary device:

```console
❯ ls -l /sys/dev/char/226:*/device/drm/
'/sys/dev/char/226:1/device/drm/':
total 0
drwxr-xr-x 13 root root 0 Sep  6 15:14 card1/
lrwxrwxrwx  1 root root 0 Sep  6 15:14 controlD65 -> card1/
drwxr-xr-x  3 root root 0 Sep  6 15:14 renderD128/

'/sys/dev/char/226:2/device/drm/':
total 0
drwxr-xr-x 8 root root 0 Sep  6 15:14 card2/
lrwxrwxrwx 1 root root 0 Sep  6 15:14 controlD66 -> card2/
drwxr-xr-x 3 root root 0 Sep  6 15:14 renderD129/
```

src/node/mod.rs
chrisduerr: I cannot speak to the motivation of the current implementation, but perhaps the goal was to guarantee the device exists on Linux? It doesn't seem *wrong* at least.

src/node/mod.rs
chrisduerr: Nope, configs are unchanged.

src/node/mod.rs
chrisduerr: Yeah this is my bad I forgot what it was doing half-way through implementing it.

src/node/mod.rs
ids1024: `drmNodeIsDRM` in libdrm also has a special case for Linux, though I guess it's different: https://gitlab.freedesktop.org/mesa/drm/-/blob/main/xf86drm.c?ref_type=heads#L3328-3334

Presumably there's some reason.

src/node/mod.rs
ids1024: The comment is a relic from before https://github.com/Smithay/smithay/commit/b098caa4579 when it did take a `RawFd`.

src/node/mod.rs
MarijnS95: Perhaps use `|` instead of `+`?

src/node/mod.rs
MarijnS95: Whoops I misread `"openbsd" != "freebsd"`, never mind.

EDIT: Again, cannot "Mark as resolved" my own review comments.

src/node/mod.rs
MarijnS95: Seeing as the same lifetime comment wasn't on the `A: AsRawFd` function there, I think we're good to remove it as this is already communicated by the API.

src/node/mod.rs
MarijnS95: Git blame tells us the explanation:

https://gitlab.freedesktop.org/mesa/drm/-/commit/f8392583418aef5e27bfed9989aeb601e20cc96d

Can we have a comment to that effect (asking generically for most special-casing here). I'm even totally happy with a one-line summary and then linking to upstream `libdrm` commits - especially if the code was based on `libdrm` :grin:

src/node/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+    ))?;
+
+    for entry in read.flatten() {
+        let name = entry.file_name();
+        let name = name.to_string_lossy();
+
+        // Only 1 primary, control and render node may exist simultaneously, so the
+        // first occurrence is good enough.
+        if name.starts_with(ty.minor_name_prefix()) {
+            let path = [r"/", "dev", "dri", &name].iter().collect::<PathBuf>();
```
MarijnS95: Also wouldn't it be cleaner to write `Path::new("/dev/dri").join(&name)`? right now this code is already specifying the separator for the root, and there's probably more allocations and work going on in general to append this.


A `format!()`ting with `/dev/dri/{}{}` is happening in other places too.

src/node/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+            "Could not find node of type {} from DRM device {}:{}",
+            ty,
+            major(dev),
+            minor(dev)
+        ),
+    ))
+}
+
+#[cfg(target_os = "openbsd")]
+fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
```
MarijnS95: For consistency with the others?

```suggestion

/// Returns the path of a specific type of node from the DRM device described by major and minor device numbers.
#[cfg(target_os = "openbsd")]
fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
```

src/node/mod.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    /// Creates a DRM node from a file stat.
+    pub fn from_stat(stat: Stat) -> Result<DrmNode, CreateDrmNodeError> {
+        let dev = stat.st_rdev;
+        DrmNode::from_dev_id(dev)
+    }
+
+    /// Creates a DRM node from a [`dev_t`].
+    pub fn from_dev_id(dev: dev_t) -> Result<DrmNode, CreateDrmNodeError> {
```
MarijnS95: Should we turn on `clippy::use_self`?

```suggestion
    pub fn from_dev_id(dev: dev_t) -> Result<Self, CreateDrmNodeError> {
```

src/node/mod.rs
MarijnS95: Thanks! Was there still any reason to define `get_minor_base()` instead of (inlining or..) going via `node_type as u32 << 6`?

You already have an explicit match (rather than this `as u32`) and documentation in `DrmNode::from_dev_id()`, perhaps those can be generalized as associated functions on `NodeType`?

src/node/mod.rs
chrisduerr: Why? I would say `+` is easier to understand for readers and both are correct in this case.

src/node/mod.rs
chrisduerr: Yeah using pathbufs here directly is much better.

src/node/mod.rs
chrisduerr: Honestly I think it's fine either way. I personally prefer `Self` because it makes text-based refactorings easier, but realistically most people are going to use LSP for stuff like that anyway so it rarely **really** matters.

src/node/mod.rs
chrisduerr: The code is used in two places so a separate function seemed appropriate.

I do not want to rely on enum layout since that's insanely easy to screw up, however there's no reason why it couldn't be a function on `NodeType` so I've moved it there.

src/node/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+        return Err(io::Error::new(
+            ErrorKind::NotFound,
+            format!("{}:{} is no DRM device", major(dev), minor(dev)),
+        ));
+    }
+
+    let old_id = minor(dev);
+    let id_mask = 0b11_1111;
+    let id = old_id & id_mask + ty.minor_base();
+    let path = PathBuf::from("/dev/dri", ty.minor_name_prefix(), id);
```
MarijnS95: Is this missing a `.join()` and/or `format!()` call now?

src/node/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    if let Some(dev_name) = devname(dev) {
+        let suffix = dev_name.trim_start_matches(|c: char| !c.is_numeric());
+        if let Ok(old_id) = suffix.parse::<u32>() {
+            let id_mask = 0b11_1111;
+            let id = old_id & id_mask + ty.minor_base();
+            let path = PathBuf::new("/dev/dri")
+                .join(ty.minor_name_prefix())
+                .join(id);
```
MarijnS95: Note that this woild result in `/dev/dri/card/0` rather than `/dev/dri/card0`. We still need to format the final name, unfortunately (not sure if there's a `.push_path()` like equivalent that doesn't insert a separator).

src/node/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+    ))?;
+
+    for entry in read.flatten() {
+        let name = entry.file_name();
+        let name = name.to_string_lossy();
+
+        // Only 1 primary, control and render node may exist simultaneously, so the
+        // first occurrence is good enough.
+        if name.starts_with(ty.minor_name_prefix()) {
+            let path = Path::new("/dev/dri").join(&*name);
```
MarijnS95, line 297: Is `&*` needed if we pass the original non-`String`?

src/node/mod.rs
MarijnS95: > however there's no reason why it couldn't be a function on `NodeType` so I've moved it there.

Yeah that's what I wanted to move to (either by returning `0, 64, 128` or `0, 1, 2` and `<< 6`'ing it whenever necessary).

src/node/mod.rs
chrisduerr: Yeah no clue what I was smoking here.

src/node/mod.rs
chrisduerr: Yeah I clearly wasn't awake yet when I wrote this, I thought there was a `/` between the arguments.

I think it's best to just go back to the full format call for these.

src/node/mod.rs
chrisduerr: Yeah it's a cow so we need this.

chrisduerr:
@MarijnS95 Is there anything still missing from this? Considering that this code was already part of the Smithay project before this PR, I was hoping this wouldn't take weeks to merge. If you're not interested in shipping this please let me know so I don't have to wait for this PR and can instead implement this myself in my application.

MarijnS95:
@chrisduerr I am not at all a maintainer of this crate and won't have any say in what gets merged nor when (and you'll either way have to "wait" for a release). I hope you've been seeing this review as constructive rather than unnecessary push-back.

Let's see if I can catch back up to where we left off, after leaving town since posting my last review.

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
+//! Module for abstractions on drm device nodes.
+
+pub mod constants;
+
+use std::error::Error;
+use std::fmt::Debug;
+use std::fmt::{self, Display, Formatter};
```
MarijnS95: Merge? I personally prefer qualifying `fmt::` everywhere to not import `::{self}`.

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+
+impl From<io::Error> for CreateDrmNodeError {
+    #[inline]
+    fn from(err: io::Error) -> Self {
+        CreateDrmNodeError::Io(err)
+    }
+}
+
+#[cfg(target_os = "freebsd")]
+fn devname(dev: dev_t) -> Option<String> {
```
MarijnS95, line 208: Maybe add a doc-comment that the returned path is relative to _within `devfs`_ (i.e. typically mounted to `/dev`)?

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    // SAFETY: The buffer written to by devname_r is guaranteed to be NUL terminated.
+    unsafe { dev_name.set_len(libc::strlen(buf)) };
+
+    Some(String::from_utf8(dev_name).expect("Returned device name is not valid utf8"))
+}
+
+/// Returns if the given device by major:minor pair is a drm device
+///
```
MarijnS95:
```suggestion
```

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+}
+
+impl Display for NodeType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        Debug::fmt(self, f)
+    }
+}
+
+/// An error that may occur when creating a DrmNode from a file descriptor.
```
MarijnS95:
```suggestion
/// An error that may occur when creating a [`DrmNode`] from a file descriptor.
```

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    Some(String::from_utf8(dev_name).expect("Returned device name is not valid utf8"))
+}
+
+/// Returns if the given device by major:minor pair is a drm device
+///
+#[cfg(target_os = "linux")]
+pub fn is_device_drm(dev: dev_t) -> bool {
+    // We `stat` the path rather than comparing the major to support dynamic device numbers:
+    //   https://gitlab.freedesktop.org/mesa/drm/-/commit/f8392583418aef5e27bfed9989aeb601e20cc96d
```
MarijnS95, line 238: Thanks for the details! Just unsure what the extra whitespace is for:

```suggestion
    // https://gitlab.freedesktop.org/mesa/drm/-/commit/f8392583418aef5e27bfed9989aeb601e20cc96d
```

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+        return None;
+    }
+
+    // SAFETY: The buffer written to by devname_r is guaranteed to be NUL terminated.
+    unsafe { dev_name.set_len(libc::strlen(buf)) };
+
+    Some(String::from_utf8(dev_name).expect("Returned device name is not valid utf8"))
+}
+
+/// Returns if the given device by major:minor pair is a drm device
```
MarijnS95: Here and below to match the rest of your doc-comments:
```suggestion
/// Returns if the given device by `major:minor` pair is a DRM device.
```

But since this is not an accurate representation (it's just a number), switch that to `the DRM device described by major and minor device numbers` that you're using in other doc-comments?

```suggestion
/// Returns if the given device described by major and minor device numbers is a DRM device.
```

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+///
+#[cfg(target_os = "linux")]
+pub fn is_device_drm(dev: dev_t) -> bool {
+    // We `stat` the path rather than comparing the major to support dynamic device numbers:
+    //   https://gitlab.freedesktop.org/mesa/drm/-/commit/f8392583418aef5e27bfed9989aeb601e20cc96d
+    let path = format!("/sys/dev/char/{}:{}/device/drm", major(dev), minor(dev));
+    stat(path.as_str()).is_ok()
+}
+
+/// Returns if the given device by major:minor pair is a drm device
```
MarijnS95:
```suggestion
/// Returns if the given device described by major and minor device numbers is a DRM device.
```

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+pub fn is_device_drm(dev: dev_t) -> bool {
+    devname(dev).map_or(false, |dev_name| {
+        dev_name.starts_with("drm/")
+            || dev_name.starts_with("dri/card")
+            || dev_name.starts_with("dri/control")
+            || dev_name.starts_with("dri/renderD")
+    })
+}
+
+/// Returns if the given device by major:minor pair is a drm device
```
MarijnS95:
```suggestion
/// Returns if the given device described by major and minor device numbers is a DRM device.
```

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+/// Returns the path of a specific type of node from the same DRM device as an existing [`DrmNode`].
+pub fn node_path(node: &DrmNode, ty: NodeType) -> io::Result<PathBuf> {
+    dev_path(node.dev, ty)
+}
+
+/// Returns the path of a specific type of node from the DRM device described by major and minor device numbers.
+#[cfg(target_os = "linux")]
+pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
```
MarijnS95, line 273: We should probably link the same "dynamic device numbers" reason for special-casing this.

At least the commit upstream that introduced `drmGetMinorNameForFD()` (`dev_path()` equivalent) was merged before that fix to `drmNodeIsDRM()`, and it's not given any details: https://gitlab.freedesktop.org/mesa/drm/-/commit/0ca03a4087a550646de7f26b6b53a932e8546474

As a side-question, at this point I wonder if it is useful to have `#[doc(alias = "drmNodeIsDRM")]` etc to point to the "identical" upstream `libdrm` function?

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            "Could not find node of type {} from DRM device {}:{}",
+            ty,
+            major(dev),
+            minor(dev)
+        ),
+    ))
+}
+
+/// Returns the path of a specific type of node from the DRM device described by major and minor device numbers.
+#[cfg(target_os = "openbsd")]
```
MarijnS95: Just checking, this should probably be:

```suggestion
#[cfg(not(any(target_os = "linux", target_os = "freebsd")))]
```

To match the `#else` in `libdrm`.

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            if path.exists() {
+                return Ok(path);
+            }
+        }
+    }
+
+    Err(io::Error::new(
+        ErrorKind::NotFound,
+        format!(
+            "Could not find node of type {} from DRM device {}:{}",
```
MarijnS95, line 307: Here and below:
```suggestion
            "Could not find node of type {} for DRM device {}:{}",
```

src/node/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            minor(dev)
+        ),
+    ))
+}
+
+/// Returns the path of a specific type of node from the DRM device described by major and minor device numbers.
+#[cfg(target_os = "freebsd")]
+fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
+    // Based on libdrm `drmGetMinorNameForFD`. Should be updated if the code
+    // there is replaced with anything more sensible...
```
MarijnS95, line 319: Yeah quite surprising that FreeBSD doesn't use the generic path with `major()` and `minor()`, those are at least used to form the error message.

Here a `doc(alias)` may be nice too.

src/node/mod.rs
MarijnS95: I think I meant when passing the original `entry.file_name()`, but fine since the conversion happens either way.

MarijnS95 (approved):
Looking nice! I pretty much only have straightforward suggestions to make the documentation in this module more consistent with itself, but the implementations look sensible now!

src/node/mod.rs
chrisduerr: I don't think there's anything wrong with the current way.

src/node/mod.rs
chrisduerr: I don't think this is necessary.

src/node/mod.rs
chrisduerr: We're not `libdrm` there's no reason why we'd try to match it. New platforms should be tested before accepting an existing impl as correct.

src/node/mod.rs
chrisduerr: It's not a doc comment and I don't think an alias here provides much value.

chrisduerr:
> I hope you've been seeing this review as constructive rather than unnecessary push-back.

Just for the record, I do not see this review as constructive. I didn't even write any of the code in Smithay and much of it was maybe not ideal, but fine as-is. Changing things around from one style to another in the same repository is just pointless code churn that risks introducing new bugs with no benefit.

Don't get me wrong I agree with many of your points, but effectively you've just stalled out what should be a "free win" for everyone into a weeks-long back and forth. For the future I would recommend just making all your suggestions in one review and then getting things resolved quickly while you are still willing to review the PR. If you don't have the time to take another look, just dismiss the review and move on, the code is fine as-is anyway.

src/node/mod.rs
i509VCB: The return of this value is probably a bit more dynamic than statically typed.

Also interesting that you now see control nodes in the DRM device folders?

MarijnS95:
> Changing things around from one style to another in the same repository is just pointless code churn that risks introducing new bugs with no benefit.

I disagree here: the fact that code was once part of **a different repository** under the same *organization* doesn't suddenly mean that it's ready for this public crate, as is evident from all the silly and unnecessary inconsistencies that I had to point out. As a user of _just the `drm-rs` crate_ that's all _I_ care about, not the rest of the Smithay organizations nor codebases.
I've spent quite some time and effort in the past to start cleaning up the style and fixing lacking doc-comments in `drm-rs` - though eventually bailed on this massive yet thankless effort.

It'd be unfortunate to have to add this module to that effort, though on the other hand is a valid way to speed up merging of a new (and apparently desperately needed?) feature, as long as there is commitment to fix i.e. documentation comments in the very near future (by the reviewer that requested them).

> you've just stalled out what should be a "free win" for everyone into a weeks-long back and forth.

Sounds like you missed my previous comment. I'm not a maintainer here and wouldn't in any way have been able to help you speed up merging - not to mention pushing out a new release. Given that no maintainer has shown up to say something along the lines of "dismiss/ignore his review comments" they're either unavailable or waiting for my review to pan out?

> For the future I would recommend just making all your suggestions in one review and then getting things resolved quickly

This comment pops up far too often. We're all human, and I don't see everything during the first round of review, or find that my review comments caused a misunderstanding when they are addressed.

> just dismiss the review and move on

Notice that I already _approved_ your PR in the second round, with only minor inconsistencies pointed out (i.e. `drm` being capitalized only half the time in doc-comments).

Please remove the re-request for review if you don't care about it. Permissions make me unable to _unassign myself_ :disappointed: 

> the code is fine as-is anyway.

Says the submitter :grimacing:

src/node/mod.rs
MarijnS95: Then how about doing the same for `is_device_drm()`?

src/node/mod.rs
MarijnS95: > It's not a doc comment

here -> above the function definition that this inner comment is describing.

src/node/mod.rs
MarijnS95: It's just inconsistent with some other functions that take a `dev: dev_t`, that's all.

src/node/mod.rs
Drakulix: Given this isn't even a public function, not sure this is necessary? Wouldn't hurt though, but I don't think this needs to block merging.

src/node/mod.rs
Drakulix: > We should probably link the same "dynamic device numbers" reason for special-casing this.

We did in other places and just repeat the code here. I would also be fine with a link/mention of `is_drm_device`.

src/node/mod.rs
Drakulix: Would apprechiate consistency here either way. Either all special cased methods should be available on a new platform or none. And I tend to agree that new platforms should be explicitly added.

Drakulix:
As a maintainer, I feel like some comments on the situation are necessary (though I apprechiate the rather honest, but at least respectful discourse so far).

> Changing things around from one style to another in the same repository is just pointless code churn that risks introducing new bugs with no benefit.

I don't think most of the review changes (especially consistency / style nits) are posing much risk here.

> Don't get me wrong I agree with many of your points, but effectively you've just stalled out what should be a "free win" for everyone into a weeks-long back and forth.

I don't think we necessarily needed this "free win". The code in smithay works and will until whenever this PR is resolved. smithay doesn't hugely benefit from this and drm-rs is a different crate with a bunch of different properties and demands (different maintainers, different regards to api stability, etc).

The reason this was merged into smithay in it's current state wasn't, that it was deemed good code. It was deemed better than what we had previously, which is generally a quite low bar to clear and mostly why smithay is much more active than drm-rs is. Both approaches make sense in regard to the particular project they are employed within, but they don't translate from one to the other just because of the shared organisation.

However I do also see, that this causes friction and dissatisfaction with the review process and obviously I do welcome these changes quite a bit. I think `DrmNode` is a great fit for drm-rs. New code should clear a certain bar of consistency and code quality, but the way I see it @chrisduerr has done his due diligence on the matter.


On specifically resolving this PR: I feel like the only comments that still need to be addressed before merging are those that affect the public api, which mostly comes down to consistency of `#[cfg]`-annotations.

The doc comments are fine in their current state and can be polished up later, if somebody feels like doing that. I'll give the code a second review round before merging, but it looks correct and style discussions around `format!` vs `PathBuf::join` or the like don't seem very productive to me at this point either. Again everybody is free to clean this up, if they feel like it, but this is nothing that needs to block merging at this point imo.

MarijnS95:
> and style discussions around `format!` vs `PathBuf::join` or the like don't seem very productive to me at this point either.

It looks like all such discussions have already been resolved in the weeks prior, in particular the semantic difference between `.join(a).join(b)` vs `format!("{}{}", a, b)` (the latter has no path separator). But now that you bring it up, it does look like `let path = Path::new("/dev/dri").join(&*name);` could be a `PathBuf::from(format!("/dev/dri/{name}"))` to be consistent with the other two implementations of `fn dev_path()` :grin: 

---

Other than that, I don't want to say that the original code from Smithay was _bad_ or anything like that. Just that we now have to opportunity to go over it once more and pick out little inconsistencies which tend to be easier to clear up directly when new code is introduced.

chrisduerr:
I think everything that needed to be said here has been said, but maybe just a few closing thoughts from me:

>>  you've just stalled out what should be a "free win" for everyone into a weeks-long back and forth.
>
> I'm not a maintainer here and wouldn't in any way have been able to help you speed up merging [...].
> Given that no maintainer has shown up [...], they're either unavailable or waiting for my review to pan out?

If I have a frequent contributor submit a review for a PR as a maintainer, I think it's logical to take a step back and wait for these things to be resolved. So if that review slows down, that can obviously still impact the time it takes to merge things.

>> the code is fine as-is anyway.
>
> Says the submitter 😬

To be fair I'm submitter, not author. The original code was completely unchanged. I do think that the quality of the original code was bad, but my goal was obviously to get this available in a reasonable timeframe so I didn't want to open the whole can of worms that a rewrite can bring (as you can see here).

> I don't think we necessarily needed this "free win". The code in smithay works and will until whenever this PR is resolved. 

It's a free win for me, because I also want to use this code. :smile: 

Generally speaking I don't think the contents of the review were a particularly big issue. Most of the comments are valid and I'd only consider some minor stuff as too nitpicky in the context of not having authored this code myself.

My main problem is that there was no review for 2 weeks. You're obviously free to work or not whenever you feel like it and you quickly made another review once I pinged you, but if you're unavailable and not the one responsible for merging anyway, it's very unfortunate to stall out on inactivity due to that.

src/node/mod.rs
chrisduerr: I wasn't aware that we already used this in other places. Consistency with other code in the module makes sense to me.

MarijnS95:
> so I didn't want to open the whole can of worms that a rewrite can bring (as you can see here).

A complete rewrite wasn't and isn't needed at all, just many smaller touch-ups, clarifications, and comment improvements is all.

> My main problem is that there was no review for 2 weeks. You're obviously free to work or not whenever you feel like it and you quickly made another review once I pinged you, but if you're unavailable and not the one responsible for merging anyway, it's very unfortunate to stall out on inactivity due to that.

I don't think maintainers should have to be waiting for "unrelated contributors" to re-review a PR, unless there has been an explicit agreement that they will be furthering it and want to stay in the loop, as well as have some final say before the merge button is clicked.

The thing about reviewing like that is that there is no obligation for me to do anything, nor in any specific time frame. At the same time that means there is no guarantee that I won't miss out on future changes and pushes before a maintainer deems it ready and merges: that's the point of not having any responsibility as a non-maintainer 😬

---

And personally I don't think 2 weeks is anything unreasonable. This PR wasn't opened with a time frame or dependency requirement in mind: to quote, it's just a nicety "to make it easier for other consumers to reuse this code". I have a lot of "this would be nice to have" PRs open (the only two other open PRs on this repository currently are good examples) that sat stale for way longer, and I might rebase or bump them every once in a while when I get back to a project. I'd love for that list to be shorter, but that's both down to maintainer/community time, and my own time to keep following up. At least there's visibility that something exists.

Drakulix:
> My main problem is that there was no review for 2 weeks. You're obviously free to work or not whenever you feel like it and you quickly made another review once I pinged you, but if you're unavailable and not the one responsible for merging anyway, it's very unfortunate to stall out on inactivity due to that.

Yeah sorry, I had like a major sickness period followed by some very much needed holidays in the last two weeks, and this just got lost in the mountains of github notifications I was slowly catching up on at the end of last week.

Which is why I very much appreciated the ping on the matter. In general, if I am not answering, it usually means I am actually not available, given I very rarely postpone these kind of things. But I am now also very aware I could have communicated this better.

Drakulix (approved):
Thanks a bunch for working on this. LGTM!

chrisduerr:
@Drakulix Seems like the latest unicode-width doesn't build. Do you want me to fix that in this PR or can you merge without clippy?

MarijnS95 (approved):

Drakulix:
> @Drakulix Seems like the latest unicode-width doesn't build. Do you want me to fix that in this PR or can you merge without clippy?

I would like to have a working state on master, so I am not going to merge this before that is fixed. But if you don't feel like taking care of this, I'll just do that sometime tomorrow, no worries.

ids1024:
I guess the main downside to merging this in drm-rs rather than Smithay is that drm-rs generally should have a more stable API. But I think this part of Smithay hasn't really had many API changes. So it seems quite reasonable to move here.

Otherwise it definitely seems good to have this in drm-rs. And rather expected, since it's functionality that is also found in libdrm.

chrisduerr:
> I would like to have a working state on master, so I am not going to merge this before that is fixed. But if you don't feel like taking care of this, I'll just do that sometime tomorrow, no worries.

Just to be clear, master is not in a working state currently, this PR changes nothing about that. I'll look into it later today.

drm-ffi/Cargo.toml
```diff
@@ -5,7 +5,7 @@ repository = "https://github.com/Smithay/drm-rs"
 version = "0.8.0"
 license = "MIT"
 authors = ["Tyler Slabinski <tslabinski@slabity.net>"]
-rust-version = "1.65"
+rust-version = "1.66"
```
chrisduerr, line 8: I've decided to bump rust-version everywhere, since CI does not distinguish between per-crate Rust versions. It shouldn't make a big difference and this way one can be certain that the MSRV is actually correct.

PR #204, opened 2024-10-17
freebsd: include missing libc crate

amshafer:
This causes compilation issues with the libc references in the devname_r definition in the devname function. With this compiling on freebsd succeeds.

Also make the non-linux versions of dev_path public, so other OS types can call it.

Here's the error:
```
error[E0433]: failed to resolve: use of undeclared crate or module `libc`
   --> src/node/mod.rs:215:9
    |
215 |         libc::devname_r(
    |         ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
   --> src/node/mod.rs:217:13
    |
217 |             libc::S_IFCHR, // Must be S_IFCHR or S_IFBLK
    |             ^^^^ use of undeclared crate or module `libc`

error[E0433]: failed to resolve: use of undeclared crate or module `libc`
   --> src/node/mod.rs:229:31
    |
229 |     unsafe { dev_name.set_len(libc::strlen(buf)) };
    |                               ^^^^ use of undeclared crate or module `libc`
```

Fix is relatively simple although I'm not sure if you want me to limit the libc dependency in `Cargo.toml` to freebsd as well?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "drm"
 description = "Safe, low-level bindings to the Direct Rendering Manager API"
 repository = "https://github.com/Smithay/drm-rs"
-version = "0.12.0"
+version = "0.14.0"
 license = "MIT"
 authors = ["Tyler Slabinski <tslabinski@slabity.net>", "Victoria Brekenfeld <crates-io@drakulix.de>"]
 exclude = [".gitignore", ".github"]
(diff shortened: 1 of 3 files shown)


Drakulix:
@amshafer Sorry for accidentally closing this, lets continue here https://github.com/Smithay/drm-rs/pull/205

PR #205, opened 2024-10-24
freebsd: include missing libc crate

Drakulix:
This causes compilation issues with the libc references in the devname_r definition in the devname function. With this compiling on freebsd succeeds.

Also make the non-linux versions of dev_path public, so other OS types can call it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ drm-ffi = { path = "drm-ffi", version = "0.9.0" }
 drm-fourcc = "^2.2.0"
 rustix = { version = "0.38.22", features = ["mm", "fs"] }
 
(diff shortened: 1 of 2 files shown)


Drakulix:
@amshafer Given this is an edition 2021 crate, we shouldn't need the `extern crate` and yes, I would like to limit the dependency to freebsd.

I pushed a new commit, that should already include all these changes, can you confirm, that this also fixes the problem for you? `cargo check --target x86_64-unknown-freebsd` seems to agree on my system, but I don't wanna setup a cross-compiler or VM or properly test this small change, if not necessary.

If so, I can merge this and do a patch release quickly.

amshafer:
Thanks, can confirm this does still fix the issue.

Drakulix:
Should be published as 0.14.1

Issue #207, opened 2024-10-26
Missing xf86drm.h bindings


jlHertel:
First of all thanks for the wonderful library.

When generating the bindings, bindgen is only looking at `drm.h` and `drm_mode.h` leaving `xf86drm.h` out. Most of the functions defined in `xf86drm.h` are just IOCTL calls, so I guess it made sense to not directly import them and instead implement the IOCTLs in Rust.
There is one function called `drmGetDevices2()` which isn't just plain IOCTL calls, but rather does quite some processing to list the DRM devices. I have a use-case in which I would like to use this function from Rust, but since the binding is missing, I'm unable to do.
I tried to understand the bindings configuration and propose a PR, but they are just mind-blowing to me.

Is it possible to generate bindings only for that specific function?

Thanks in advance.

ids1024:
I don't think we actually link `libdrm`. So that's why the bindings only includes ioctl numbers, structs, and such, and not `extern` functions.

It may be good to include an implementation of similar functionality, if we don't already. https://github.com/Smithay/drm-rs/pull/202 moved some code related to dealing with device nodes to drm-rs from Smithay, but I guess it doesn't include a simple way to get a list of devices.

jlHertel:
If that is the case, then we can probably port the code directly from libdrm, since its also MIT licensed.
One question though: we generally don't provide a Card implementation, but this function returns a list of cards, so we will need to provide at least a file-backed Card. Is that OK? 
Sorry if the question is dumb, I'm quite new to Rust and might be overlooking something

jlHertel:
@ids1024 I made a pull request with the code I believe will be enough to list devices. I would appreciate if you could review it.

PR #208, opened 2024-10-27
Add function to enumerate devices

jlHertel:
Adds a basic DRM Listing functionality.
Please note that, different from libDRM, we don't process the devices to include bus information or anything else. We also don't de-duplicate devices like libDRM does, so duplicates might occur.

This change addresses #207


src/node/mod.rs
```diff
@@ -380,3 +380,21 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         ),
     ))
 }
+
+/// Returns a Vector with all DRM Nodes we managed to find. There might be duplicates.
+pub fn drm_get_devices() -> Result<Vec<DrmNode>, io::Error> {
```
notgull: Could this instead return `impl Iterator<Item = io::Result<DrmNode>>`?

src/node/mod.rs
```diff
@@ -380,3 +380,21 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
...
+    let mut devices: Vec<DrmNode> = vec![];
+
+    fs::read_dir("/dev")?
+        .for_each(|entry| {
+            if entry.is_ok() {
+                let device_path = entry.unwrap().path();
+                let node = DrmNode::from_path(device_path);
+                if node.is_ok() {
+                    devices.push(node.unwrap())
+                }
```
notgull:
```suggestion
                if let Ok(node) = node {
                    devices.push(node)
                }
```

src/node/mod.rs
```diff
@@ -380,3 +380,21 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         ),
     ))
 }
+
+/// Returns a Vector with all DRM Nodes we managed to find. There might be duplicates.
+pub fn drm_get_devices() -> Result<Vec<DrmNode>, io::Error> {
+    let mut devices: Vec<DrmNode> = vec![];
```
MarijnS95: Type shouldn't be needed here.

src/node/mod.rs
```diff
@@ -380,3 +380,21 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         ),
     ))
 }
+
+/// Returns a Vector with all DRM Nodes we managed to find. There might be duplicates.
```
MarijnS95:
```suggestion
/// Returns a [`Vec`] with all DRM Nodes we managed to find. There might be duplicates.
```

src/node/mod.rs
```diff
@@ -380,3 +380,21 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
...
+pub fn drm_get_devices() -> Result<Vec<DrmNode>, io::Error> {
+    let mut devices: Vec<DrmNode> = vec![];
+
+    fs::read_dir("/dev")?
+        .for_each(|entry| {
```
MarijnS95: How about making this a `.map().collect()` rather than a side-effecting `for_each()`? If that, use regular control flow with `for entry in fs::read_dir(..)? { .. }`.

The `.collect()` can be dropped if using @notgull's suggestion to return an iterator instead.

MarijnS95 (changes requested):

src/node/mod.rs
jlHertel: Wouldn't we need to filter the results in some way if we use `.map()`? Asking as we check all devices and likely a lot of them aren't DRM devices, so we need to skip them. This validation is done right now by calling `DrmNode::from_path()`.

src/node/mod.rs
jlHertel: I'm not sure what would be the best approach here.
We can have IO Errors only by failing to open/read the `/dev` directory.
For the devices themselves, we will only get errors if they aren't DRM devices, which should be the case for the majority of stuff in `/dev`.

Would iterator improve anything in regards to the error handling?

Also, sorry if my answers are a bit confusing. I'm quite new to Rust, so I might be missing the big-picture on how to better handle errors/iterators.

src/node/mod.rs
jlHertel: Removed as suggested.

src/node/mod.rs
MarijnS95: Such filtering can be implemented by a [`.filter_map()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map), but I'm hesitant to `stat()` every device on the system to see if it is a DRM device. 

Since `drmGetDevices2()` was requested, I expected this feature to be built after that function. While it does also [appear to ignore any error from processing individual devices](https://gitlab.freedesktop.org/mesa/drm/-/blob/e68e9b8013f0cb5630afd156d330dc9665d9d98f/xf86drm.c#L4789-4791), this function [does open `DRM_DIR_NAME`](https://gitlab.freedesktop.org/mesa/drm/-/blob/e68e9b8013f0cb5630afd156d330dc9665d9d98f/xf86drm.c#L4783) which is [`/dev/dri` on every platform except FreeBSD where it is `/dev`](https://gitlab.freedesktop.org/mesa/drm/-/blob/e68e9b8013f0cb5630afd156d330dc9665d9d98f/xf86drm.h#L79-90). This way we don't have to `stat()` devices that we know won't be DRM devices in the first place.

src/node/mod.rs
MarijnS95: > We can have IO Errors only by failing to open/read the `/dev` directory.

That could be handled with:

```rust
io::Result<impl Iterator<Item = io::Result<DrmNode>>>
```

> For the devices themselves, we will only get errors if they aren't DRM devices, which should be the case for the majority of stuff in `/dev`.

And as just mentioned below, `drmGetDevices2()` skips errors while opening sub-devices, so we might as well return an iterator over `DrmNode` directly:

```rust
io::Result<impl Iterator<Item = DrmNode>>
```

> For the devices themselves, we will only get errors if they aren't DRM devices, which should be the case for the majority of stuff in `/dev`.

And as below, on most platforms except FreeBSD we should be reading `/dev/dri`.

> Also, sorry if my answers are a bit confusing. I'm quite new to Rust, so I might be missing the big-picture on how to better handle errors/iterators.

No worries, I think everyone is happy to educate new contributors on Rust as well as describing the "Rust style" used in drm-rs (as much as we're still discovering what that's supposed to be).

---

Regarding the `drm_get_devices()` function name (based on `drmGetDevices(2)()`), I'd drop the `drm_` prefix since we're already in the `drm` crate (i.e. a fully qualified call would look like `drm::node::drm_get_devices()` which is somewhat superfluous) and then also drop the `get_` prefix because of the [no `get_` prefix convention](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

src/node/mod.rs
jlHertel: Hey thanks for the pointers. I somehow completely missed the fact that `DRM_DIR_NAME` was different in certain platforms. I will update the code to also use `/dev/dri` for all platforms except FreeBSD.

src/node/mod.rs
jlHertel: Thanks for all the feedback, it's highly appreciated.
I pushed the changes with everything you suggested. Do you mind to give another look?

jlHertel:
BUMP. @MarijnS95 @notgull Can you guys give another look in this PR and approve/merge if its good?

src/node/mod.rs
```diff
@@ -380,3 +380,12 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         ),
     ))
 }
+
+/// Returns an iterator with all DRM Nodes we managed to find. There might be duplicates.
```
MarijnS95: Nit: not sure about the wording "we managed to find". We might want to be more explicit about _where_/how we found them instead, and plainly state that these are "all the DRM nodes" currently on the system (putting trust in our implementation).

jlHertel:
Checking once again if there any update on this.

drm-ffi/drm-sys/src/bindings.rs
```diff
@@ -1,4 +1,4 @@
-/* automatically generated by rust-bindgen 0.69.4 */
+/* automatically generated by rust-bindgen 0.69.5 */
```
Drakulix: Can you drop this likely accidental and automatic change?

Drakulix (approved):
LGTM apart from a slight nitpick.

src/node/constants.rs
```diff
@@ -43,3 +43,11 @@ pub const RENDER_NAME: &str = "renderD";
...
 #[cfg(target_os = "openbsd")]
 pub const RENDER_NAME: &str = "drmR";
+
+/// Directory to search DRM devices
+#[cfg(target_os = "openbsd")]
+pub const DRM_DIR_NAME: &str = "/dev";
+
+/// Directory to search DRM devices
+#[cfg(not(target_os = "openbsd"))]
+pub const DRM_DIR_NAME: &str = "/dev/dri";
```
MarijnS95, line 53: Should we use these constants in more places in `node/mod.rs`?

src/node/mod.rs
```diff
@@ -4,9 +4,9 @@ pub mod constants;
 
 use std::error::Error;
 use std::fmt::{self, Debug, Display, Formatter};
-use std::io;
 use std::os::unix::io::AsFd;
 use std::path::{Path, PathBuf};
+use std::{fs, io};
```
MarijnS95: We should probably merge all module imports, or move `fmt::{self` here and flatten that a bit.

src/node/mod.rs
```diff
@@ -380,3 +380,12 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
         ),
     ))
 }
+
+/// Returns an iterator with all DRM Nodes we managed to find. There might be duplicates.
+pub fn devices() -> io::Result<impl Iterator<Item = DrmNode>> {
+    let result = fs::read_dir(DRM_DIR_NAME)?
+        .filter_map(|entry| entry.ok())
+        .filter_map(|entry| DrmNode::from_path(entry.path()).ok());
```
MarijnS95, line 388: Nit: I'd only filter out `from_path() == Err(CreateDrmNodeError::NotDrmNode)`, and bubble up all other `io::Error`s to the user.

This would quite literally look like `fs::read_dir()` that returns an _outer and inner_ `io` error: `-> io::Result<impl Iterator<Item = io::Result<DrmNode>>>`.

And maybe all `entry.ok()`s should translate into that as well? Though OTOH, we cannot be sure if the entry we tried to visit was even DRM node in the first place, making things confusing (equally confusing as _missing_ DRM nodes because we were not able/allowed to visit that entry for some reason).

src/node/constants.rs
jlHertel: The other place we use the path `/dev/dri` is for the function `dev_path` which is our equivalent to `drmGetMinorNameForFD`. Curiously the libdrm's function uses the constant everywhere. I will update the code to also make use of it.

src/node/mod.rs
jlHertel: libdrm's `drmGetDevice2()` doesn't seem to bubble-up IO errors ([here I guess][1]) after it's reading the directory, so I'm inclined to also not bubble those up, so we remain closer to their implementation.

Of course that is my personal taste and I'm fine to change to your suggested approach if that is desired.

[1]: https://github.com/tobiasjakobi/libdrm/blob/exynos/xf86drm.c#L4145

drm-ffi/drm-sys/src/bindings.rs
jlHertel: Removed as suggested.

src/node/mod.rs
jlHertel: I dropped the "we managed to find" part.
Shall I expand the description to mention the exact directory and how we find the devices?

src/node/mod.rs
jlHertel: I moved `fmt` down to this import line. Please take another look.

src/node/mod.rs
MarijnS95: We don't need to necessarily be bit-for-bit identical to what `drmGetDeviceS2()` does, since the underlying `process_device()` returns `-1` both for errors as well as detecting it's not a DRM node.

If we do such a thing we do have to be careful indeed; it'd be a bummer to bail on an `io::Error` from `stat()` when reading a file that isn't a DRM device in the first place (i.e. on `openbsd` that reads all of `/dev`). On the other hand, it might be "annoying" to not be alerted of certain IO failures (possibly permission denied?) for paths that are DRM devices that won't get returned.

This is why I suggested returning nested errors so that the caller can decide whether to skip them or not, but I'd like to leave it to the other drm-rs maintainers to make a decision.

If anything it might be useful to look at existing users of `drm-rs` and see if they can benefit from this addition; and if so how they currently implement the looping.

MarijnS95 (approved):
I'm okay with this whichever way it falls. Given that the constants are `pub` and the implementation is trivial enough, folks can copy if they are interested in individual per-`entry` errors.

src/node/mod.rs
MarijnS95: I think I should have been more elaborate here (but wasn't, since it's not related to this PR): personally I don't like importing _both_ a module and _some_ of its items. In this case `fmt` is only used in the `impl` blocks for `fmt::Result`, but `Debug`/`Display`/`Formatter` are globally imported.

If that's to disambiguate the `Result` from the prelude, my personal preference is to import only `fmt` and qualify `fmt::Debug`/`fmt::Display`/`fmt::Formatter` :)

src/node/mod.rs
MarijnS95: Oh also, looking at the examples and I specifically just opened #218, how about updating all the examples in this repo/crate to use this function in favour of the hardcoded `/dev/dri/card0`? I had to hack up a bunch of examples previously when using an Intel GPU which seemed to start probing at `/dev/dri/card1`.

jlHertel: I had another look at the examples and realized that in libdrm we open the devices and thus have a file descriptor ready. In this library devices aren't opened and no node/file path information is stored, so even if we are able to list all devices, we can't do much with them, as we don't know how to open them.

I guess it might make sense to extend the current implementation to also include at least the path of the device, otherwise it won't be very useful for the listing.

PR #210, opened 2024-11-01
node: Fix mask-or operator precedence by replacing `+` with `|`

### LINKED ISSUE #209: Fails to detect DRM nodes on FreeBSD
Related to #202. Sorry, I don't know Rust to debug more.

```
$ su root -c "rm /usr/local/lib/dri/swrast_dri.so" # prevent silent fallback
$ vidcontrol -s 2 </dev/ttyv0 # same as Ctrl+Alt+F2
$ cosmic-comp
[...]
ERROR cosmic_comp::backend::kms Failed to initialize software EGL renderer. err=no EGL device found with `EGL_MESA_device_software`
[...]

$ COSMIC_RENDER_DEVICE=/dev/dri/renderD128 cosmic-comp
<works fine>
```

Reverting to [Smithay version](https://github.com/Smithay/smithay/blob/e8af9a1e45f6/src/backend/drm/node/mod.rs#L312-L324) appears to help:
```diff
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -330,8 +330,18 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     if let Some(dev_name) = devname(dev) {
         let suffix = dev_name.trim_start_matches(|c: char| !c.is_numeric());
         if let Ok(old_id) = suffix.parse::<u32>() {
-            let id_mask = 0b11_1111;
-            let id = old_id & id_mask + ty.minor_base();
+            let old_ty = match old_id >> 6 {
+                0 => NodeType::Primary,
+                1 => NodeType::Control,
+                2 => NodeType::Render,
+                _ => {
+                    return Err(io::Error::new(
+                        ErrorKind::NotFound,
+                        format!("{}:{} is no DRM device", major(dev), minor(dev)),
+                    ));
+                }
+            };
+            let id = old_id - old_ty.minor_base() + ty.minor_base();
             let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
             if path.exists() {
                 return Ok(path);
```

MarijnS95:
Fixes #209, CC @jbeich

A request for using masks and shifts on the known `0-1-2` constants defining the minor type of a DRM node resulted in the expression to maintain the original `+` from first subtracting the current node type followed by adding the requested node type.  As the subtract of the current node type was replaced by a mask, the resulting `old_id & ID_MASK + minor_base` expression is interpreted as `old_id & (ID_MASK + minor_base)` because of operator precedence, while it would have been correctly interpreted as `(old_id & ID_MASK) | minor_base` when the correct `|` OR operator is used.

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -330,8 +330,8 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     if let Some(dev_name) = devname(dev) {
         let suffix = dev_name.trim_start_matches(|c: char| !c.is_numeric());
         if let Ok(old_id) = suffix.parse::<u32>() {
-            let id_mask = 0b11_1111;
-            let id = old_id & id_mask + ty.minor_base();
+            const ID_MASK: u32 = 0b11_1111;
+            let id = old_id & ID_MASK | ty.minor_base();
             let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
             if path.exists() {
                 return Ok(path);
@@ -363,8 +363,8 @@ pub fn dev_path(dev: dev_t, ty: NodeType) -> io::Result<PathBuf> {
     }
 
     let old_id = minor(dev);
-    let id_mask = 0b11_1111;
-    let id = old_id & id_mask + ty.minor_base();
+    const ID_MASK: u32 = 0b11_1111;
+    let id = old_id & ID_MASK | ty.minor_base();
     let path = PathBuf::from(format!("/dev/dri/{}{}", ty.minor_name_prefix(), id));
     if path.exists() {
         return Ok(path);
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -44,16 +44,7 @@ impl DrmNode {
             return Err(CreateDrmNodeError::NotDrmNode);
         }
 
-        // The type of the DRM node is determined by the minor number ranges:
-        //   0 -  63 -> Primary
-        //  64 - 127 -> Control
-        // 128 - 255 -> Render
-        let ty = match minor(dev) >> 6 {
-            0 => NodeType::Primary,
-            1 => NodeType::Control,
-            2 => NodeType::Render,
-            _ => return Err(CreateDrmNodeError::NotDrmNode),
-        };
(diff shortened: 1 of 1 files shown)


MarijnS95:
CI says:

> error: package `fdeflate v0.3.6` cannot be built because it requires rustc 1.67.0 or newer, while the currently active rustc version is 1.66.0

Should we bump MSRV, lock `fdeflate` to an older version, downgrade `fdeflate` in CI [like we already do for `home`](https://github.com/Smithay/drm-rs/blob/b65efbc210db58aa5d8b2878006dbc82335fdda3/.github/workflows/ci.yml#L83-L84) or run the MSRV tests with `cargo +nightly -Zminimal-versions generate-lockfile`? I prefer the last one because that's what I do in many other crates. It not only exercises `-Zminimal-versions` which downstream crates/consumers may also be running, but also prevents this error from showing up in the future if any other crate bumps their MSRV beyond our `rust-version` in a patch release.

Drakulix:
> [...] or run the MSRV tests with `cargo -nightly -Zminimal-versions generate-lockfile`?

I think we do want to test against the newest available and compiler supported version though. We have a separate minimal test...

MarijnS95:
> I think we do want to test against the newest available and compiler supported version though. We have a separate minimal test...

Specifically, you want to *check MSRV against the latest crate dependencies*? That means we'll always be at the merit of our (transitive) dependencies bumping MSRV, and our CI subsequently failing on random PRs.

I'd prefer to combine the MSRV test with the `minimal-versions` test to keep overhead low. Right now the existing `cargo check` runs on `nightly` as well, the suggested change generates the lockfile on `nightly` while running `check` on the MSRV compiler (and we could also check it on the `stable` compiler).

MarijnS95:
Note that that's what I did in https://github.com/Smithay/drm-rs/pull/212, and we should probably continue this discussion there. Alternatively, we perhaps can and should exclude our examples / `dev-dependencies` from the MSRV build?

MarijnS95:
@Drakulix I've also included all CI fixes in this one PR (the oldest) so that it can hopefully all go in **using rebase-merge** while the CI is green/happy with the final result :)

Drakulix (approved):
LGTM, thanks!

PR #211, opened 2024-11-05
CI: Replace deprecated/unmaintained `actions-rs` with simple `run: steps`

MarijnS95:
The `actions-rs` containers haven't been maintained and updated for years and don't need to: GitHub's actions environment already comes fully loaded with a complete `stable` Rust installation with the standard tools (clippy, rustfmt, rustdoc).  Simple `run:` commands relate directly to what a developer can type in locally to "reproduce" the CI environment while they might be following up on CI failures, and no longer spam ancient Node 12 deprecation warnings. Whenever a different toolchain or additional targets are needed, https://github.com/dtolnay/rust-toolchain is used to install it instead.

The only downside is that `actions-rs/clippy-check` properly embeds build failures as code annotations in the diff; but as shown in e.g. https://github.com/Smithay/drm-rs/actions/runs/11632990715/job/32397210203 this doesn't even work for PRs from forks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,14 +13,6 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
-      - name: Setup Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          components: rustfmt
-          default: true
-          override: true
       - name: Cargo cache
         uses: actions/cache@v3
         with:
@@ -29,23 +21,12 @@ jobs:
             ~/.cargo/git
           key: ${{ runner.os }}-cargo-rust_stable-${{ hashFiles('**/Cargo.toml') }}
       - name: Format
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
+        run: cargo fmt --all -- --check
 
   doc:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
-      - name: Setup Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: nightly
-          profile: minimal
-          components: rust-docs
-          default: true
-          override: true
       - name: Cargo cache
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -54,11 +35,7 @@ jobs:
             ~/.cargo/git
           key: ${{ runner.os }}-cargo-rust_nightly-${{ hashFiles('**/Cargo.toml') }}
       - name: Documentation
-        uses: actions-rs/cargo@v1
-        env:
-          DOCS_RS: 1
```
MarijnS95, line 59: We don't seem to rely on https://docs.rs/about/builds#detecting-docsrs, maybe one of our dependencies does?

.github/workflows/ci.yml
```diff
@@ -69,17 +46,7 @@ jobs:
           sudo apt-get install -y libdrm-dev
       - name: Fetch drm headers
         run: ./.github/workflows/fetch_headers.sh
-      - name: Update deps
-        uses: actions-rs/cargo@v1
-        with:
-          command: update
```
MarijnS95: Not sure what this was intended for: no `Cargo.lock` is checked in so a fresh checkout will inevitably regenerate a fresh lockfile on the latest version.

.github/workflows/ci.yml
```diff
@@ -95,10 +62,7 @@ jobs:
           path: target
           key: ${{ runner.os }}-build-rust_1.66.0-check-${{ hashFiles('**/Cargo.toml') }}
       - name: Clippy
-        uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all --all-features --all-targets -- -D warnings -A clippy::redundant_static_lifetimes
+        run: cargo clippy --all --all-features --all-targets -- -D warnings -A clippy::redundant_static_lifetimes
```
MarijnS95, line 70: We'll loose annotations here in favour of plain text-based `cargo clippy` output, but that wasn't/isn't working for PRs from forks anyway.

Drakulix (approved):
Thanks, sorry for taking so long with this.

.github/workflows/ci.yml
MarijnS95: Oh I see. I remember from long ago that `-Zminimal-versions` on at least `cargo check` does *nothing* when a `Cargo.lock` is already available. Hence, by running this `cargo update` first we **never perform a minimum-versions check**. You can see this on the latest CI build of the `develop` branch where only latest versions (i.e. `proc-macro2 v1.0.95`) are used:

https://github.com/Smithay/drm-rs/actions/runs/16524679480/job/46734820812

On the CI in this PR, you can see it actually downgraded to `proc-macro2 v1.0.55`:

https://github.com/Smithay/drm-rs/actions/runs/16524859270/job/46735448393?pr=211

And then things fail, because the minimum versions are set "too loosely" :grin:

Drakulix:
Merged via #227

PR #212, opened 2024-11-05
CI: Combine MSRV-check with `-Zminimal-versions` build

MarijnS95:
Sometimes dependency crates bump their MSRV beyond our MSRV in a patch release, and fail our CI in subsequent versions.  Instead of downgrading independent crates in CI every time this issue arises, combine it with our already-existing `-Zminimal-versions` check.  Instead of running this check on `nightly`, we only use that toolchain to generate a `Cargo.lock` and subsequently `check` the code on our MSRV toolchain.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,10 +69,17 @@ jobs:
           sudo apt-get install -y libdrm-dev
       - name: Fetch drm headers
         run: ./.github/workflows/fetch_headers.sh
-      - name: Update deps
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+          default: true
+          override: true
+      - name: Generate minimal-versions lockfile
         uses: actions-rs/cargo@v1
         with:
-          command: update
+          command: generate-lockfile
+          args: -Z minimal-versions
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: 1.66.0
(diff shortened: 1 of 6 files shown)


MarijnS95:
https://github.com/Smithay/drm-rs/actions/runs/11681253865/job/32525926098?pr=212

```
error[E0433]: failed to resolve: use of undeclared crate or module `syntax`
Error:  --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/regex-1.5.0/src/literal/mod.rs:9:9
  |
9 |     use syntax::hir::literal::Literals;
  |         ^^^^^^ use of undeclared crate or module `syntax`
```

It's been a while since I last saw this one: https://github.com/rust-lang/regex/issues/931. It's showing up now because the original MSRV check builds with `--all-features`, which we now combine with a `-Zminimal-versions` build as I outlined in https://github.com/rust-lang/regex/issues/931#issuecomment-1885251793. Bumping our minimum `bindgen` version to `0.69.3` which includes https://github.com/rust-lang/regex/issues/931 should solve this. Otherwise we have to change our `--all --all-features` build?

MarijnS95:
Note that the original `-Zminimal-versions` build wasn't building minimal versions at all. See a detailed breakdown at https://github.com/Smithay/drm-rs/pull/211#discussion_r2231333701.

Drakulix:
Superseeded by #211 & #227

PR #216, opened 2025-02-25
Update OpenBSD support to new drm API introduced in 6.9

tobhe:
This makes OpenBSD work properly on my machine.

WIP because it depends on a libc change in https://github.com/rust-lang/libc/pull/4285#issuecomment-2682420476

My rust experience is rather limited so I'm looking forward to hear how this could be improved.

One potential problem I see is that `devname()` is not thread-safe as is because the name buffer is static. Does this need to be copied safely, guarded by a lock or marked unsafe?

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -86,7 +86,7 @@ impl DrmNode {
 
     /// Returns whether the DRM device has render nodes.
     pub fn has_render(&self) -> bool {
-        #[cfg(target_os = "linux")]
+        #[cfg(any(target_os = "linux", target_os = "openbsd"))]
         {
             node_path(self, NodeType::Render).is_ok()
         }
@@ -98,7 +98,7 @@ impl DrmNode {
             false
         }
 
-        #[cfg(not(any(target_os = "linux", target_os = "freebsd")))]
+        #[cfg(not(any(target_os = "linux", target_os = "freebsd", target_os = "openbsd")))]
         {
             false
         }
--- a/src/node/constants.rs
+++ b/src/node/constants.rs
@@ -21,25 +21,10 @@ pub const DRM_MAJOR: u32 = 87;
 pub const DRM_MAJOR: u32 = 226;
 
 /// Primary DRM node prefix.
-#[cfg(not(target_os = "openbsd"))]
 pub const PRIMARY_NAME: &str = "card";
 
-/// Primary DRM node prefix.
-#[cfg(target_os = "openbsd")]
-pub const PRIMARY_NAME: &str = "drm";
-
 /// Control DRM node prefix.
-#[cfg(not(target_os = "openbsd"))]
 pub const CONTROL_NAME: &str = "controlD";
 
-/// Control DRM node prefix.
-#[cfg(target_os = "openbsd")]
-pub const CONTROL_NAME: &str = "drmC";
-
 /// Render DRM node prefix.
-#[cfg(not(target_os = "openbsd"))]
 pub const RENDER_NAME: &str = "renderD";
-
-/// Render DRM node prefix.
-#[cfg(target_os = "openbsd")]
-pub const RENDER_NAME: &str = "drmR";


MarijnS95:
This was based on upstream `libdrm`, which still defines those original constants for OpenBSD to be different from the rest: https://gitlab.freedesktop.org/mesa/drm/-/blob/a7eb2cfd53a70fcd9ba9dcfad80a3994642f362f/xf86drm.h#L79-90

Any clue as to how/when this was changing, and is this PR backwards-compatible "enough"?

src/node/mod.rs
```diff
@@ -231,6 +231,27 @@ fn devname(dev: dev_t) -> Option<String> {
...
+    use std::ffi::{CStr};
+
+    let buf: *mut c_char = unsafe {
+        libc::devname(
+            dev,
+            libc::S_IFCHR, // Must be S_IFCHR or S_IFBLK
+        )
+    };
+
+    // Buffer was too small (weird issue with the size of buffer) or the device could not be named.
```
MarijnS95: That's a stale comment copied from above, since you didn't allocate a "too small" buffer yourself :)

src/node/mod.rs
```diff
@@ -231,6 +231,27 @@ fn devname(dev: dev_t) -> Option<String> {
     Some(String::from_utf8(dev_name).expect("Returned device name is not valid utf8"))
 }
 
+#[cfg(target_os = "openbsd")]
+fn devname(dev: dev_t) -> Option<String> {
+    use std::os::raw::{c_char, c_int};
+    use std::ffi::{CStr};
+
+    let buf: *mut c_char = unsafe {
```
MarijnS95: No need for an explicit type when it's defined by the API:

```suggestion
    let buf = unsafe {
```

tobhe:
> Any clue as to how/when this was changing, and is this PR backwards-compatible "enough"?

It looks like this changed in 2021 https://github.com/openbsd/xenocara/commit/9d1e1e287e6f15bd4810a2651a500918a3c54d54

OpenBSD only supports the last two releases, with one release every 6 months that means we don't need to worry (besides, breaking API and ABI across releases is not a huge issue in OpenBSD land).

src/node/mod.rs
tobhe: Literally copied this from the non-openbsd version above but I am happy to change it.

src/node/mod.rs
tobhe: Indeed

tobhe:
Latest push includes a few fixes based on feedback from @MarijnS95. I also added a check to handle the actual error case "??" properly. I am not convinced that `buf.is_null()` is actually possible here but it probably doesn't hurt to keep that check.

Drakulix:
> One potential problem I see is that devname() is not thread-safe as is because the name buffer is static. Does this need to be copied safely, guarded by a lock or marked unsafe?

We can't really mark a function as thread-unsafe in rust, so I am afraid this would need a global lock variable.

tobhe:
Clarified the `/dev/dri` node change in the commit message, changed the libc dependency to a more realistic target (if I understand their versioning right) and added a global lock for the static buffer returned by `devname()`.

tobhe:
libc with `devname` support is now available so I force pushed to remove the XXX comment and removed WIP in the PR title

src/node/mod.rs
```diff
@@ -231,6 +231,40 @@ fn devname(dev: dev_t) -> Option<String> {
...
+            libc::S_IFCHR, // Must be S_IFCHR or S_IFBLK
+        )
+    };
+
+    if buf.is_null() {
+        return None;
+    }
+
+    let dev_str =
+        unsafe { CStr::from_ptr(buf).to_str() }.expect("Returned device name is not valid utf8");
```
MarijnS95: I think `to_str()` can be outside of the `unsafe` block where `.expect()` already resides.

src/node/mod.rs
```diff
@@ -231,6 +231,40 @@ fn devname(dev: dev_t) -> Option<String> {
     Some(String::from_utf8(dev_name).expect("Returned device name is not valid utf8"))
 }
 
+#[cfg(target_os = "openbsd")]
+use std::sync::Mutex;
+
+#[cfg(target_os = "openbsd")]
+static DEVNAME_LOCK: Mutex<()> = Mutex::new(());
```
MarijnS95: Would it make sense to keep this `static` locally inside the function for now? And/or do we expect other calls to `devname()`?

Note that this won't help if user code - through other means - calls `libc::devname()` without knowing about "your / our" lock 😅

MarijnS95 (approved):
Nice! Wish we could retitle this PR to explain that it's not really a "fix" but a "change" to accomodate for recent breaking changes in the OpenBSD API that you linked?

src/node/mod.rs
MarijnS95: Feel free to clean that one up too.

src/node/mod.rs
tobhe: > Would it make sense to keep this static locally inside the function for now? And/or do we expect other calls to devname()?

Sure, I can do that

> Note that this won't help if user code - through other means - calls libc::devname() without knowing about "your / our" lock 😅

True, but that's how the underlying API works so not much we can do about it imho.

Drakulix: While I suggested the lock, I wonder if we should instead mark `devname` unsafe and all functions using it, while documenting that `this is unsound to call from multiple threads on some platforms`.

tobhe:
I finally found some time to look into this again and as it turns out some of the recent fixes also helped with OpenBSD. In particular we were also affected by https://github.com/Smithay/drm-rs/commit/e854a2b92016042f1dbd17fbbec392101a0a16ba and after further research into libdrm I think we want to stick to this implementation for `dev_path` on OpenBSD which makes the entire `devname` dance redundant (luckily).

With that the only things missing are fixing the node names and i found that we can also enable `has_render()` on OpenBSD.

PR #217, opened 2025-03-06
[API Break] Make `control::property::Info::value_type` return `&` instead of clone

ids1024:
It seems odd for this to return an owned value, when `::name()` doesn't. And it is useful to get a reference to the `ValueType`, so `ValueType::convert_value` can be used and just use the lifetime of the `Info` instead of the cloned `ValueType.`

I assume there's no reason not to do this, since callers can still clone as needed.

As a small API break, I think this should be merged whenever the next semver bump to the crate happens to be.

### DIFF
--- a/examples/planes.rs
+++ b/examples/planes.rs
@@ -24,7 +24,7 @@ fn display_plane(
     for (handle, value) in &props {
         let info = card.get_property(*handle)?;
         let name = info.name().to_str().unwrap().to_owned();
-        prop_map.insert(name, (info.value_type(), *value));
+        prop_map.insert(name, (info.value_type().clone(), *value));
     }
     let Value::Enum(Some(type_)) = convert_value(&prop_map["type"]) else {
         panic!("failed to convert plane type enum");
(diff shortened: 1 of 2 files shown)


MarijnS95:
It was probably intended as a cheap Copy type, except that the presence of EnumValues (with Vecs) no longer makes that cheaply possible?

ids1024:
Yeah, I was wondering about that, but haven't check the history. If it were a `Copy` type the current API would be good, but not now that it contains a couple `Vec`s (if it's an enum), and `Value<'_>` holds a reference to it.

Drakulix:
I think this is a good change, is this ready for merge?

ids1024:
I think the only reason I marked this as a draft is because it's a breaking change. So if there's going to be a breaking release, it should be mergable.

MarijnS95 (approved):

MarijnS95:
Didn't we already land a bunch of breaking changes? This can probably be merged after a rebase, I don't see the CI being unhappy with it after that :)

Drakulix:
yeah, I think this just needs a rebase

ids1024:
CI passing, after also updating a line in `planes` example.

Drakulix:
thx!

PR #218, opened 2025-03-06
Add `planes` example

ids1024:
This shows information about DRM planes, under the CRTC and connector they are associated with. I've found this useful for trying to understand what's going on with planes vs trying to read `drm_info` output, so it may as well be added here.


Drakulix:
Could you rebase this PR?

Drakulix (approved):
thx!

PR #221, opened 2025-05-13
node: Fix behavior on DragonFlyBSD

ids1024:
DragonFlyBSD is similar to FreeBSD, so it should use the FreeBSD implementation here instead of the default implementation (which I guess is really just the NetBSD/OpenBSD implementation).

This now works properly, at least with this simple test:

```rust
use std::fs;

fn main() {
    for i in fs::read_dir("/dev/dri").unwrap() {
        let i = i.unwrap();
        let node = drm::node::DrmNode::from_path(i.path()).unwrap();
        println!("{}: {:?}", i.path().display(), node);
    }
}
```


Drakulix (approved):

ids1024:
This can also be merged before the next release, though it's not anything hugely important.

PR #222, opened 2025-05-14
Fix clippy lints from latest Rust version

ids1024:


Drakulix:
I guess we need to downgrade `flate2` as well..

Drakulix:
Fixed with https://github.com/Smithay/drm-rs/pull/225

PR #223, opened 2025-05-18
update to rustix 1.0 and bump linux-raw-sys

m4rch3n1ng:
the only really relevant changes were the changes to the ioctl module, replacing the type generic with a const generic and the types with consts. see the last section of [rustix's `CHANGES.md`](https://github.com/bytecodealliance/rustix/blob/92466d9860d95aa15b8ad0a5ea4cb88b39ccc63f/CHANGES.md). the only other change was in an example where `-1` (meaning no timeout) was replaced with `None` (also meaning no timeout).

i also updated the `linux-raw-sys` dependency of `drm-sys` to match the version required by `rustix` 1, to avoid duplicating that dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,14 +15,14 @@ bytemuck = { version = "1.14", features = ["extern_crate_alloc", "derive"] }
 bytemuck_derive = { version = "1.5" }
 drm-ffi = { path = "drm-ffi", version = "0.9.0" }
 drm-fourcc = "^2.2.0"
-rustix = { version = "0.38.22", features = ["mm", "fs"] }
+rustix = { version = "1.0.7", features = ["mm", "fs"] }
 
 [target.'cfg(target_os = "freebsd")'.dependencies]
 libc = "0.2"
 
 [dev-dependencies]
 image = { version = "0.24", default-features = false, features = ["png"] }
-rustix = { version = "0.38.22", features = ["event", "mm"] }
+rustix = { version = "1.0.7", features = ["event", "mm"] }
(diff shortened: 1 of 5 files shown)


Drakulix (approved):

dextero:
Hi folks!

I'm also interested in using drm.rs with rustix 1.0. Can I help push this forward? :)

I found that the MSRV check failure can be worked around by either:

- Pinning more transitive dependencies [as it's done for `home` right now](https://github.com/Smithay/drm-rs/blob/develop/.github/workflows/ci.yml#L84):
  ```
  cargo update -p fdeflate@0.3.7 --precise 0.3.5
  cargo update -p flate2@1.1.2 --precise 1.0.35
  ```
- Removing the dependency on `home` by updating [`bindgen` build dep](https://github.com/Smithay/drm-rs/blob/develop/drm-ffi/drm-sys/Cargo.toml#L18) to 0.70+ and disabling the [`with-dirs`](https://github.com/kkawakam/rustyline/blob/master/Cargo.toml#L79) feature of `rustyline`. This unfortunately requires bumping MSRV to 1.70, but removes the need for explicitly downgrading `home` in CI.

If either of those sounds like an acceptable solution I'll gladly make a PR.

Drakulix:
Sorry for getting back to this so late, but could you rebase your PR? CI was fixed on `develop` and I would like to get your changes into the next release. Thanks!

m4rch3n1ng:
done

Drakulix (approved):
Thanks! :)

PR #225, opened 2025-07-25
Minimal CI fixes

MarijnS95:
Excerpt from #210 with just the _minimal_ CI fixes, that we can rebase-merge to enable all other PRs to be merged on top again.

After that we can incrementally improve the existing CI steps again.


Drakulix (approved):
Thanks a lot!

PR #226, opened 2025-07-25
syncobj: Fix import_sync_file and add timeline exports/imports

### LINKED ISSUE #224: syncobj::fd_to_handle with import_sync_file = true is wrong
DRM_SYNCOBJ_FD_TO_HANDLE_FLAGS_IMPORT_SYNC_FILE imports a fence fd into an *existing* syncobj handle which needs to be passed into `handle`. But the current function always passes `handle: 0` on line 73 so it cannot work: https://github.com/Smithay/drm-rs/blob/b65efbc210db58aa5d8b2878006dbc82335fdda3/drm-ffi/src/syncobj.rs#L67-L88

The API should have a separate function like:
```rust
pub fn syncobj_import_sync_file(
    fd: BorrowedFd<'_>,
    handle: u32,
    sync_file: BorrowedFd<'_>,
) -> io::Result<drm_syncobj_handle> {
    let mut args = drm_syncobj_handle {
        handle,
        flags: DRM_SYNCOBJ_FD_TO_HANDLE_FLAGS_IMPORT_SYNC_FILE,
        fd: sync_file.as_raw_fd(),
        pad: 0,
    };

    unsafe {
        ioctl::syncobj::fd_to_handle(fd, &mut args)?;
    }

    Ok(args)
}
```
(idk if the return value is any useful here)

Drakulix:
This addresses the broken api reported in #224 and also provides entry point for the new optional timeline argument.

Ping @ids1024, @YaLTeR, @MarijnS95

### DIFF
--- a/drm-ffi/src/syncobj.rs
+++ b/drm-ffi/src/syncobj.rs
@@ -7,10 +7,7 @@ use drm_sys::*;
 
 use std::{
     io,
-    os::{
-        fd::{FromRawFd, OwnedFd},
-        unix::io::{AsRawFd, BorrowedFd},
-    },
+    os::fd::{AsRawFd, BorrowedFd},
 };
 
 /// Creates a syncobj.
@@ -43,7 +40,7 @@ pub fn destroy(fd: BorrowedFd<'_>, handle: u32) -> io::Result<drm_syncobj_destro
 }
 
 /// Exports a syncobj as an inter-process file descriptor.
-pub fn handle_to_fd(fd: BorrowedFd<'_>, handle: u32) -> io::Result<OwnedFd> {
+pub fn handle_to_fd(fd: BorrowedFd<'_>, handle: u32) -> io::Result<drm_syncobj_handle> {
     let mut args = drm_syncobj_handle {
         handle,
         flags: 0,
@@ -56,31 +53,40 @@ pub fn handle_to_fd(fd: BorrowedFd<'_>, handle: u32) -> io::Result<OwnedFd> {
         ioctl::syncobj::handle_to_fd(fd, &mut args)?;
     }
 
-    Ok(unsafe { OwnedFd::from_raw_fd(args.fd) })
+    Ok(args)
 }
 
-/// Exports a syncobj as a poll-able sync file.
+/// Exports a syncobj as a poll-able `sync_file` optionally for a specified timeline point.
 pub fn handle_to_sync_file(
     fd: BorrowedFd<'_>,
     handle: u32,
-) -> io::Result<OwnedFd> {
+    timeline_point: Option<u64>,
+) -> io::Result<drm_syncobj_handle> {
     let mut args = drm_syncobj_handle {
         handle,
-        flags: DRM_SYNCOBJ_HANDLE_TO_FD_FLAGS_EXPORT_SYNC_FILE,
+        flags: DRM_SYNCOBJ_HANDLE_TO_FD_FLAGS_EXPORT_SYNC_FILE
+            | if timeline_point.is_some() {
+                DRM_SYNCOBJ_HANDLE_TO_FD_FLAGS_TIMELINE
+            } else {
+                0
+            },
         fd: 0,
         pad: 0,
-        point: 0, // TODO: Add support for TIMELINE sync files
+        point: timeline_point.unwrap_or(0),
     };
 
     unsafe {
         ioctl::syncobj::handle_to_fd(fd, &mut args)?;
     }
 
-    Ok(unsafe { OwnedFd::from_raw_fd(args.fd) })
+    Ok(args)
 }
 
-/// Imports a file descriptor exported by [`handle_to_fd`] back into a process-local handle.
-pub fn fd_to_handle(fd: BorrowedFd<'_>, opaque_fd: BorrowedFd<'_>) -> io::Result<u32> {
+/// Imports a file descriptor exported by [`handle_to_fd()`] back into a process-local handle.
+pub fn fd_to_handle(
+    fd: BorrowedFd<'_>,
+    opaque_fd: BorrowedFd<'_>,
+) -> io::Result<drm_syncobj_handle> {
     let mut args = drm_syncobj_handle {
         handle: 0,
         flags: 0,
@@ -93,28 +99,35 @@ pub fn fd_to_handle(fd: BorrowedFd<'_>, opaque_fd: BorrowedFd<'_>) -> io::Result
         ioctl::syncobj::fd_to_handle(fd, &mut args)?;
     }
 
-    Ok(args.handle)
+    Ok(args)
 }
 
-/// Imports a sync file exported by [`handle_to_sync_file`] into an existing process-local handle.
+/// Imports a `sync_file` exported by [`handle_to_sync_file()`] into an existing process-local handle,
+/// optionally for a specified timeline point.
 pub fn sync_file_into_handle(
     fd: BorrowedFd<'_>,
     syncobj_fd: BorrowedFd<'_>,
     handle: u32,
-) -> io::Result<()> {
+    timeline_point: Option<u64>,
+) -> io::Result<drm_syncobj_handle> {
     let mut args = drm_syncobj_handle {
         handle,
-        flags: DRM_SYNCOBJ_FD_TO_HANDLE_FLAGS_IMPORT_SYNC_FILE,
+        flags: DRM_SYNCOBJ_FD_TO_HANDLE_FLAGS_IMPORT_SYNC_FILE
+            | if timeline_point.is_some() {
+                DRM_SYNCOBJ_FD_TO_HANDLE_FLAGS_TIMELINE
+            } else {
+                0
+            },
         fd: syncobj_fd.as_raw_fd(),
         pad: 0,
-        point: 0, // TODO: Add support for TIMELINE sync files
+        point: timeline_point.unwrap_or(0),
     };
 
     unsafe {
         ioctl::syncobj::fd_to_handle(fd, &mut args)?;
     }
 
-    Ok(())
+    Ok(args)
 }
 
 /// Waits for one or more syncobjs to become signalled.
--- a/drm-ffi/src/syncobj.rs
+++ b/drm-ffi/src/syncobj.rs
@@ -7,7 +7,10 @@ use drm_sys::*;
 
 use std::{
     io,
-    os::unix::io::{AsRawFd, BorrowedFd},
+    os::{
+        fd::{FromRawFd, OwnedFd},
+        unix::io::{AsRawFd, BorrowedFd},
+    },
 };
 
 /// Creates a syncobj.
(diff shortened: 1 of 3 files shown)


MarijnS95:
Can we maybe have a clearer standalone title? When looking at git log, I don't have a mapping to github issues.

Will look at the code later.

Drakulix:
> Can we maybe have a clearer standalone title? When looking at git log, I don't have a mapping to github issues.

Changed the PR title and split the commits into smaller better described chunks.

src/control/mod.rs
```diff
@@ -857,13 +858,15 @@ pub trait Device: super::Device {
...
     }
 
-    /// Imports a sync file exported by [`Self::syncobj_to_sync_file`] into an existing process-local handle.
+    /// Imports a sync file exported by [`Self::syncobj_to_sync_file`] into an existing process-local handle
+    /// optionally for a specified timeline point.
     fn sync_file_into_syncobj(
         &self,
         fd: BorrowedFd<'_>,
         handle: syncobj::Handle,
+        timeline_point: Option<u64>,
```
YaLTeR: If I'm not mistaken, this timeline point is "on" the sync file `fd` (rather than on the target syncobj `handle`), so it would make more sense for this argument to be next to `fd`

drm-ffi/src/syncobj.rs
```diff
@@ -7,7 +7,10 @@ use drm_sys::*;
 
 use std::{
     io,
-    os::unix::io::{AsRawFd, BorrowedFd},
+    os::{
+        fd::{FromRawFd, OwnedFd},
+        unix::io::{AsRawFd, BorrowedFd},
```
MarijnS95: Should we just import all of these consistently from `os::fd`?

drm-ffi/src/syncobj.rs
```diff
@@ -39,54 +42,91 @@ pub fn destroy(fd: BorrowedFd<'_>, handle: u32) -> io::Result<drm_syncobj_destro
...
+    };
+
+    unsafe {
+        ioctl::syncobj::handle_to_fd(fd, &mut args)?;
+    }
+
+    Ok(unsafe { OwnedFd::from_raw_fd(args.fd) })
+}
+
+/// Exports a syncobj as a poll-able sync file optionally for a specified timeline point.
```
MarijnS95:
```suggestion
/// Exports a syncobj as a poll-able `sync_file`, optionally for a specified timeline point.
```

drm-ffi/src/syncobj.rs
```diff
@@ -39,54 +42,91 @@ pub fn destroy(fd: BorrowedFd<'_>, handle: u32) -> io::Result<drm_syncobj_destro
...
 
     unsafe {
         ioctl::syncobj::handle_to_fd(fd, &mut args)?;
     }
 
-    Ok(args)
+    Ok(unsafe { OwnedFd::from_raw_fd(args.fd) })
 }
 
 /// Imports a file descriptor exported by [`handle_to_fd`] back into a process-local handle.
```
MarijnS95:
```suggestion
/// Imports a file descriptor exported by [`handle_to_fd()`] back into a process-local handle.
```

drm-ffi/src/syncobj.rs
```diff
@@ -39,54 +42,91 @@ pub fn destroy(fd: BorrowedFd<'_>, handle: u32) -> io::Result<drm_syncobj_destro
...
+    };
+
+    unsafe {
+        ioctl::syncobj::fd_to_handle(fd, &mut args)?;
+    }
+
+    Ok(args.handle)
+}
+
+/// Imports a sync file exported by [`handle_to_sync_file`] into an existing process-local handle optionally for a specified timeline point.
```
MarijnS95:
```suggestion
/// Imports a `sync_file` exported by [`handle_to_sync_file()`] into an existing process-local handle, optionally for a specified timeline point.
```

Not sure whether `sync_file` really is the right naming, otherwise maybe `Sync File` with a link to https://docs.kernel.org/driver-api/sync_file.html?

drm-ffi/src/syncobj.rs
```diff
@@ -39,54 +42,91 @@ pub fn destroy(fd: BorrowedFd<'_>, handle: u32) -> io::Result<drm_syncobj_destro
...
+pub fn sync_file_into_handle(
```
MarijnS95, line 107: Guess you used `into` over `to` to add weight to the fact that it goes _in_to an existing handle?

drm-ffi/src/syncobj.rs
```diff
@@ -39,54 +42,91 @@ pub fn destroy(fd: BorrowedFd<'_>, handle: u32) -> io::Result<drm_syncobj_destro
...
-        point: 0, // TODO: Add support for TIMELINE sync files
+        point: timeline_point.unwrap_or(0),
     };
 
     unsafe {
         ioctl::syncobj::handle_to_fd(fd, &mut args)?;
     }
 
-    Ok(args)
+    Ok(unsafe { OwnedFd::from_raw_fd(args.fd) })
```
MarijnS95: Could you make the same change to `fn create()` and maybe other functions?

OTOH it looks intended for the `drm-ffi` crate to not reason about which files may have been modified inside the ioctl struct, which is both a consistent pattern but also makes these functions a really weird abstraction in relation to the high-level `drm` crate.

src/control/mod.rs
MarijnS95: It's on the handle. The `syncobj` can be a timeline and `sync_file` would then be the `dma_fence` that is waited on or signaled at the given timeline point.


https://docs.kernel.org/gpu/drm-mm.html#drm-sync-objects

Drakulix: good call, done

drm-ffi/src/syncobj.rs
Drakulix: I think that naming is good. and documentation should be easy to find, if you deal with linux drm.

drm-ffi/src/syncobj.rs
Drakulix: yes

src/control/mod.rs
MarijnS95: ? just making sure, it's on the `handle` so the current position makes sense right? It's not on the `fd`, which "simply" a binary `sync_file`.

drm-ffi/src/syncobj.rs
Drakulix: Yeah I am not sure, perhaps I should change them back to return the whole struct for consistency. Possibly we shouldn't even split things like `fd_to_syncobj` and `sync_file_into_syncobj` given they use the same `ioctl`. However then we have two `Option`s and incorrect combinations, making correct usage much more difficult to address.

I am not opposed to change `drm-ffi` to use some more higher-level types including `OwnedFd`, but I don't really want to do this as part of this PR.

I guess the original idea for this abstraction was to make it possible to use the calls directly if further down the line new uses get added and/or existing padded-bytes re-used etc. This is an example of that not really working given `point` was added later.

Maybe @Slabity has some opinions on this?

MarijnS95: If that was the case, perhaps now is the time to call it on that experiment and unify things back into a single `drm`/`drm-sys` crate in a separate PR.

The only alternative I can see is _also_ accepting the struct as an input, but that's already the API that `ioctl::syncobj::xxx()` takes :)

src/control/mod.rs
Drakulix: my bad. fixed

drm-ffi/src/syncobj.rs
Drakulix: I changed this PR to return the whole ioctl struct again and do the conversion in `drm-rs`. But I am happy to continue this discussion in a future PR or issue on the matter.

src/control/mod.rs
YaLTeR: Hm. This is quite confusing and poorly documented on the kernel side tbh, for example on that drm-mm page I see no mention of `DRM_SYNCOBJ_FD_TO_HANDLE_FLAGS_TIMELINE`. Guess I might've confused this call with timeline transfer. I suppose this is correct as is now.

drm-ffi/src/syncobj.rs
Slabity: So the original idea with the division of crates was to follow what was, at the time, best practices for interfacing with system libraries and calls.

- The `sys` crate would be exclusively generated bindings to the raw C calls and structures. The only logic involved here would be the build scripts and how they clean things up.
- The `ffi` crate would do the bare minimum to allow making the calls safely without doing any sort of logic involved. Such as setting up memory pointers/lengths for the various structures that use them. No assumptions other than the bare minimum required to make the `ioctl` calls are done here.
- The `rs` crate would do everything higher order and try to closely reflect what `libdrm` does, leveraging Rust's ability to divide things through traits and also provide a clean, well documented interface. Any sort of logic, dynamic allocation, or other non-trivial things would be done here.

To be honest, I don't know if that's considered best practice now. I do think a separate `ffi` crate makes things more flexible if someone doesn't like how a certain `drm-rs` call is. It's also pretty clear 90% of the code in `ffi` is basically zero cost as well. But I also don't know much about the `syncobj` API here to argue about it cleanly.

The concern about a stable interface is something that will need to be addressed eventually though. To be honest, I don't think it's possible to follow semantic versioning well in the `sys` or `ffi` crates.
- The `sys` crate is obvious because we have no control over any changes in the userspace API and we rely on the promise that "We don't break userspace".
- The `ffi` crate because technically `ioctl` calls can be extended and may require new arguments to be added, which is a breaking change.

As far as I know, it's normal to never really stabilize the `sys` or `ffi` crates. But I don't really know if the "best practice" here changed at all.

src/control/mod.rs
MarijnS95: Yeah, the commit adding it didn't extend any docs (https://patchwork.freedesktop.org/patch/646004), but it makes sense when reading the bigger definition of `syncobj` and `sync_file`, and when considering how/why a `sync_file` (multiple if you have a timeline...) are imported into an *existing* `syncobj` instead of creating a new one like the opaque file descriptors.

drm-ffi/src/syncobj.rs
MarijnS95: > I changed this PR to return the whole ioctl struct again and do the conversion in `drm-rs`.

Note that you went out of your way to document the input arguments to these functions, but didn't document what return fields of the struct are expected to be relevant - I think that's where my main disconnect with this API design comes from.

---

> * The `ffi` crate would do the bare minimum to allow making the calls safely without doing any sort of logic involved. Such as setting up memory pointers/lengths for the various structures that use them. No assumptions other than the bare minimum required to make the `ioctl` calls are done here.

Fair enough though it does all become cloudy when considering inout parameters, optional fields and such. The caller has a clear view of what goes into the function, not what comes out.

In other words, having this crate doesn't seem too bad, but returning the full `ioctl()` struct is what seems to be confusing perhaps?

---

> As far as I know, it's normal to never really stabilize the `sys` or `ffi` crates. But I don't really know if the "best practice" here changed at all.

Not sure if this is standard, but most high-level crates might still utilize `sys` types in their public API for a variety of reasons. Assuming they're not always plain `type` aliases, any breaking change to the `-sys` crate automatically means a break to the high-level crate.

Though for the crates I work on (i.e. `ndk`) those breaking changes in the underlying C API are already "disallowed" as they would also break ABI. Certain cases like enums are exempt via `#[non_exhaustive]` tricks.

Assuming that is the case for the `sys` crate here as well (adding `#[non_exhaustive]` to all structs, `ioctl()`s should be aware of size for new/optional fields?) that could very well be "stabilizable". It leaves `ffi` as "the odd one out" when having to extend function parameter interfaces.

ids1024:
Should these functions be called `import_sync_file()` and `export_sync_file()`? Given the name of the flags, the fact they're called `drmSyncobjImportSyncFile` and `drmSyncobjExportSyncFile` in `libdrm`, and in the kernel call the functions `drm_syncobj_export_sync_file()` and `drm_syncobj_import_sync_file_fence()`.

drm-ffi/src/syncobj.rs
Drakulix: Looking at this again, it doesn't seem like we reached any sort of conclusion here.

So given the original vision, we probably shouldn't split the function in drm-ffi. The only ever useful parts of the ioctl struct are the file descriptor and handle, which I guess could be returned unconditionally as `(RawFd, u32)` or something.

`drm` would then actually differentiate between the various uses of the ioctl. Sounds good?

PR #227, opened 2025-07-25
CI: Fix minimal-versions job

Drakulix:
Continuation of https://github.com/Smithay/drm-rs/pull/211

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,14 +13,6 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
-      - name: Setup Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          components: rustfmt
-          default: true
-          override: true
       - name: Cargo cache
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -69,39 +46,27 @@ jobs:
...
         with:
           path: target
-          key: ${{ runner.os }}-build-rust_1.66.0-check-${{ hashFiles('**/Cargo.toml') }}
+          key: ${{ runner.os }}-build-rust_1.70.0-check-${{ hashFiles('**/Cargo.toml') }}
       - name: Clippy
-        uses: actions-rs/clippy-check@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all --all-features --all-targets -- -D warnings -A clippy::redundant_static_lifetimes
+        run: cargo +1.70.0 clippy --all --all-features --all-targets -- -D warnings -A clippy::redundant_static_lifetimes
```
MarijnS95, line 69: I'm only noticing now that we were only running `clippy` on our MSRV and not on `stable`. Is that intended?

Perhaps reduces contributor churn by not flagging small new contributions as failing CI (on irrelevant existing hunks in this codebase) if the repo hasn't been touched for a while and there's a new `stable` with stricter lints (similar to the stuff we had to do in #225 to get it back to a functioning state).

PR #234, opened 2025-12-23
Bump all dependencies

linkmauve:
Only `rustyline` required a code change.

We might want to regenerate `drm-ffi/drm-sys/src/bindings.rs` with the newer `bindgen` at some point.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: |
+          uname -a
           sudo apt-get update -y
-          sudo apt-get install -y libdrm-dev
+          sudo apt-get install -y libdrm-dev linux-headers-6.8.0-1044-azure
       - name: Fetch drm headers
         run: ./.github/workflows/fetch_headers.sh
       - uses: dtolnay/rust-toolchain@1.89.0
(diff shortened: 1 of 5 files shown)


linkmauve:
The two checks are because the current msrv is 1.70, whereas the newer dependencies require a more recent rustc. Should I downgrade these dependencies to keep compatibility with that June 2023 version, or would it be fine to bump our msrv?

Slabity:
Sorry for the delay on reviewing these MRs. Going through a bunch of them now.

I have no requirements on MSRV as I use nightly or the latest Stable for most things. I think it should be fine to update them as anything shipped downstream would likely update their versions anyways when releasing it as part of a distribution package or something similar.

linkmauve:
Sorry, I think I broke that PR, and I don’t quite understand why.  The check pass is complaining about `linux/bits.h` not existing, which should be fixed by my last WIP commit but isn’t. The check-minimal pass I understand better, the image crate depends on png which depends on flate2 which depends on miniz-sys which depends on gcc, but all of these are in super old versions which required an unstable feature from nightly, but I don’t quite get how to fix that, unless with time traveling.

Slabity:
> Sorry, I think I broke that PR, and I don’t quite understand why. The check pass is complaining about `linux/bits.h` not existing, which should be fixed by my last WIP commit but isn’t. The check-minimal pass I understand better, the image crate depends on png which depends on flate2 which depends on miniz-sys which depends on gcc, but all of these are in super old versions which required an unstable feature from nightly, but I don’t quite get how to fix that, unless with time traveling.

No worries. It probably got broke by something else that got merged or some minimal version not being compatible.

I'll take a look over the next few days to see if I can figure out the problem.

HigherOrderLogic:
Hey, is there any progress on this?

PR #235, opened 2025-12-23
Use core instead of std when possible

linkmauve:
We were previously importing many things from `std` even though they are available in `core`, this will eventually help `no_std` support.


linkmauve:
Hi, I’ve just rebased this PR, is there anything missing to get it reviewed and merged?

Slabity:
Sorry for the delay. Looking at it now. Removing `std` is definitely one of the goals so this is definitely appreciated work.

PR #237, opened 2026-01-08
Add `is_empty` method to `AtomicModeReq`

my4ng:
This allows unnecessary `ioctl` system calls to be avoided if the request is empty. For example:

```rust
let mut req = AtomicModeReq::new();

if foo {
    req.add_raw_property(...);
}

if bar {
    req.add_raw_property(...);
}

if !req.is_empty() {
    device.atomic_commit(..., req);
}
```

### DIFF
--- a/src/control/atomic.rs
+++ b/src/control/atomic.rs
@@ -17,6 +17,11 @@ impl AtomicModeReq {
         Self::default()
     }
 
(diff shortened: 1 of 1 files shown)


Drakulix (approved):
thanks!

PR #241, opened 2026-04-02
upgrade rustyline to version 18

alexanderkjall:
based on: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/drm/debian/patches/rustyline-17.patch?ref_type=heads


Slabity:
Looks good to me.

PR #242, opened 2026-04-07
Avoid requiring two versions of `linux-raw-sys`.

jimblandy:
Update `drm-sys` to depend on `linux-raw-sys` 0.12, to avoid requiring both version 0.9 and version 0.12 of `linux-raw-sys`. Before this patch, the situation is (abridged):

```
├── drm-ffi v0.9.1 (/home/jimb/drm-rs/drm-ffi)
│   ├── drm-sys v0.8.1 (/home/jimb/drm-rs/drm-ffi/drm-sys)
│   │   └── linux-raw-sys v0.9.4
│   └── rustix v1.1.4
│       └── linux-raw-sys v0.12.1
```

Tested with `cargo clippy --workspace --all-features`.

### DIFF
--- a/drm-ffi/drm-sys/Cargo.toml
+++ b/drm-ffi/drm-sys/Cargo.toml
@@ -19,7 +19,7 @@ bindgen = { version = "0.70", optional = true }
 pkg-config = { version = "0.3.19", optional = true }
 
 [target.'cfg(any(target_os = "android", target_os = "linux"))'.dependencies]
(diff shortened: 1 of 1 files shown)


MarijnS95:
This is only because the latest `rustix` requires `0.12`; on `rustix 1.0.8` for example (which `drm-sys` allows via `^1`) they still depend on `^0.9.2`.

I can't think of a good argument either way (MSRV is unchanged for example), just adding some context to the callout.

PR #243, opened 2026-04-07
Rename macro `map_len` to `map_capacity`, to match Rust terminology.

jimblandy:
In `drm_ffi::utils`, rename the macro `map_len` to `map_capacity`, as it returns the capacity of the vector, not its length. The DRM types don't seem to use the term "length" or "len" either, so the old name doesn't match either Rust's or DRM's conventions.

Tested with `cargo clippy --all-features --workspace`.


linkmauve:
This macro goes away in #240, so I don’t think this PR is useful.

PR #250, opened 2026-09-17
Prevent force probe for disconnected connectors when not requested

### LINKED ISSUE #248: drm_ffi::mode::get_connector incorrectly force probes when a connector is disconnected
It first does a get_connector ioctl with a dummy mode list of size 1 to get the amount of modes and such, but then when the amount of modes is 0, it will call the get_connector ioctl again with an empty mode list rather than a dummy mode list even when the force_probe argument is false. And an empty mode list tells the kernel to probe the connector. The C version of libdrm correctly handles this: https://gitlab.freedesktop.org/mesa/libdrm/-/blob/main/xf86drmMode.c?ref_type=heads#L519-522

This is causing a performance issue for Redox OS on real hardware where probing is expensive since https://gitlab.redox-os.org/redox-os/base/-/merge_requests/392. As the bootlog and console daemons will both call get_connector on every frame to check if a virtio-gpu display has been resized or a new display has been connected in the absence of a way to signal hotplug events through the DRM device itself. (Redox OS doesn't have udev to send hotplug events over)

bjorn3:
Previously when the first get_connector ioctl returned 0 modes, the second ioctl call would pass a list of 0 modes even when force_probe is false. An empty list of modes indicates a request to force probe. With this change instead we use a dummy list of 1 mode when 0 are reported and force_probe is false.

Fixes https://github.com/Smithay/drm-rs/issues/248


bjorn3:
CI failure is unrelated to this PR. A cargo version older than necessary for a dependency is used.

